PR #9, opened 2019-04-23
BACKLOG-10130 add displayName for the form field

Taybou:


src/main/java/org/jahia/modules/contenteditor/api/forms/EditorFormField.java
```diff
@@ -58,32 +62,37 @@ public EditorFormField(String name,
...
+            field.mandatory,
+            field.valueConstraints == null ? null : field.valueConstraints.stream()
+                .map(constraint -> new EditorFormFieldValueConstraint(constraint))
+                .collect(Collectors.toList()),
+            field.defaultValues == null ? null : new ArrayList<>(field.defaultValues),
+            field.removed,
+            field.targets == null ? null : field.targets.stream()
+                .map(target -> new EditorFormFieldTarget(target))
+                .collect(Collectors.toList()),
+            // FIXME no deep copy here
```
FBerthelot: Why not Fix this in the PR ?

Taybou: Not related to this task. We should create dedicated task to fix this one.

FBerthelot: The task is created ?

If not, I think you should create it :)

Taybou: Ok, I'll created.

Taybou: After Discussion with @cedricmoitrier:
We don't see the usage of keeping `extendedPropertyDefinition`.

We would like to hear from you @sergehuber

PR #52, opened 2019-05-23
BACKLOG-10267 : fix image size in picker and image selector

dgriffon:


src/javascript/EditPanelContainer/EditPanel/EditPanelContent/FormBuilder/SelectorTypes/SelectorTypes.js
```diff
@@ -13,7 +13,8 @@ const SelectorTypes = {
     Text: () => Text,
     RichText: () => RichText,
     Picker: options => {
-        return (options && PickerOptions[options.type]) || ContentPicker;
+        let option = options && options.find(option => option.name === 'type' && PickerOptions[option.value]);
```
Taybou: minor: const instead of let.

PR #63, opened 2019-06-04
BACKLOG-10370: add datepicker component in the Design System

FBerthelot:
I've added Storybook in order to develop properly this component


FBerthelot:
@dgriffon test is working on my machine, I've removed the unused dependency

PR #71, opened 2019-06-07
BACKLOG-10400: add date picker field icon

jkevan:


src/javascript/DesignSystem/DatePickerInput/DatePickerInput.jsx
```diff
@@ -7,6 +7,8 @@ import {Input} from '../Input';
 import {javaDateFormatToJSDF} from './date.util';
 
 import dayjs from 'dayjs';
+import IconButton from '@material-ui/core/IconButton/IconButton';
```
dgriffon:
```suggestion
import {IconButton} from '@material-ui/core';
```

PR #122, opened 2019-07-02
BACKLOG-10541 handle on select language in content editor

Taybou:


FBerthelot:
I think some test could be done here, but nevermind. I merge !

PR #131, opened 2019-07-09
BACKLOG-10524 hide action upload section

Taybou:


src/javascript/DesignSystem/PickerDialog/PickerDialog.jsx
```diff
@@ -30,15 +30,15 @@ const styles = theme => ({
...
+        justifyContent: 'flex-end'
     },
     actionButtons: {
         '& button': {
             marginRight: theme.spacing.unit * 2
         }
     },
     actionUpload: {
-        display: 'flex',
+        display: 'none',
```
FBerthelot: If display none ? why not removing the element from the dom ?

Taybou: You can follow the slack discussion , PO said it will coming soon!

Taybou: Done

PR #134, opened 2019-07-10
BACKLOG-10095 fix issue of action buttons when input text is focused

Taybou:


src/javascript/EditPanelContainer/EditPanel/EditPanelContent/FormBuilder/SelectorTypes/Text/Text.jsx
```diff
@@ -5,7 +5,7 @@ import PropTypes from 'prop-types';
 import {Input} from '../../../../../../DesignSystem/Input';
 import {FieldPropTypes} from '../../../../../FormDefinitions/FromData.proptypes';
 
-export const Text = ({field, id, editorContext, formik: {values, handleChange, handleBlur}}) => {
+export const Text = ({field, id, editorContext, formik: {values, handleChange}}) => {
```
FBerthelot: This is kindof strange to do that. Can you explain why are you doing that ?

src/javascript/EditPanelContainer/EditPanel/EditPanelContent/FormBuilder/SelectorTypes/TextArea/TextArea.jsx
```diff
@@ -10,7 +10,16 @@ export const TextAreaField = ({id, field}) => {
         <Field
             name={field.formDefinition.name}
             render={props => {
-                return <TextArea id={id} disabled={field.formDefinition.readOnly} {...props.field}/>;
+                const {name, value, onChange} = props.field;
```
FBerthelot: Why not handling onBlur again ?

src/javascript/EditPanelContainer/EditPanel/EditPanelContent/FormBuilder/SelectorTypes/Text/Text.jsx
Taybou: The question why we use onBlur on this component? is there any handling for the onBlur event?

src/javascript/EditPanelContainer/EditPanel/EditPanelContent/FormBuilder/SelectorTypes/TextArea/TextArea.jsx
Taybou: Why handling it?

src/javascript/EditPanelContainer/EditPanel/EditPanelContent/FormBuilder/SelectorTypes/Text/Text.jsx
FBerthelot: Because formik rely on it for validation purpose.

We use formik for validation, otherwise formik is not needed on this project.

Taybou: What type of validation we do today?

FBerthelot: Nothing for now.

But why suddenly remove some formik features ?

Taybou: Because we haven't used it and it causes a problem, why keep something that hasn't been used and it makes problems?

FBerthelot: The initial question is what are the problem you are trying to solve. 

Removing formik issue like that only for some field and other field continue with the initial behavior of formik is quite strange.

Why keep formik on those field if we doesn't need validation ?
I think we need to investigate deeper on this issue.

Taybou: - The problem is described well on this ticket: https://jira.jahia.org/browse/BACKLOG-10095.

- Remove formik issue like that! => We should report the other broken fields to do the same for it.

- We don't do any validation for now. The day we need or wish to do validation we can use onChange for this purpose for example or we implement our custom onBlur function, and of course we have different type of controls where we can do the validation: https://jaredpalmer.com/formik/docs/guides/validation#when-does-validation-run

FBerthelot: Ok, I see.

Can you tell me what is done in the handleBlur function that avoid to trigger onClick on an other field ?

I think we should, in this specific case, understand the root cause of this bug, so we can figure out how to not reproduce this bug as long new field is added to CE.

PR #158, opened 2019-07-25
BACKLOG-10710 handle tag separator to add tags

Taybou:


src/javascript/EditPanelContainer/EditPanel/EditPanelContent/FormBuilder/SelectorTypes/Tag/Tag.jsx
```diff
@@ -37,9 +40,23 @@ const Tag = ({field, id, t}) => {
                             /* Do Nothing on blur BACKLOG-10095 */
                         }}
                         onChange={selection => {
+                            let adaptedSelection = [];
```
dgriffon: Please use named variable instead of single char.
Move this code in an utility module and provide a test for it

PR #179, opened 2019-08-07
BACKLOG-10818 add logger when it's impossible to serialize object

Taybou:


src/main/java/org/jahia/modules/contenteditor/api/forms/impl/StaticDefinitionsRegistry.java
```diff
@@ -82,14 +83,20 @@ private void registerStaticEditorFormFieldSets(Bundle bundle) {
...
+
+                if (StringUtils.isNotBlank(name)) {
+                    SortedSet<EditorFormFieldSet> editorFormFieldSets = staticEditorFormFieldSetsByName.get(name);
+                    if (editorFormFieldSets == null) {
+                        editorFormFieldSets = new TreeSet<>();
+                    }
+                    editorFormFieldSets.add(editorFormFieldSet);
+                    staticEditorFormFieldSetsByName.put(name, editorFormFieldSets);
+                    bundleEditorFormFieldSets.add(editorFormFieldSet);
+                    logger.info("Successfully loaded static form for name {} from {}", name, editorFormURL);
```
dgriffon: Please change to
```suggestion
                    logger.info("Successfully loaded static fieldSets for name {} from {}", name, editorFormURL);
```

PR #182, opened 2019-08-09
BACKLOG-10845 hide fieldSets that have "jmix:templateMixin" as mixin

Taybou:


dgriffon:
It could have been nice to use overrides but has the are not supporting inheritance we have to other way to hide such kind of type.

PR #238, opened 2019-09-19
BACKLOG-11109: preview on displayable node

jkevan:


Taybou:
Very nice work @jkevan ^^'

PR #256, opened 2019-09-25
Backlog 11170 : Setup Husky

ffffffelix:


.gitignore
```diff
@@ -15,3 +15,6 @@ graphql.config.json
 graphql.schema.json
 coverage
 /storybook-static
+
+# macOS
+.DS_Store
```
FBerthelot: This should be in the global git configuration, it's not specific to the project environement

.gitignore
```diff
@@ -15,3 +15,6 @@ graphql.config.json
 graphql.schema.json
 coverage
 /storybook-static
+
```
FBerthelot:
```suggestion
```

.gitignore
```diff
@@ -15,3 +15,6 @@ graphql.config.json
...
+# macOS
```
FBerthelot:
```suggestion
```

.gitignore
FBerthelot:
```suggestion
```

PR #280, opened 2019-10-09
BACKLOG-11309: fix create mode regression

jkevan:


src/javascript/Create/CreateForm/create.request.js
```diff
@@ -18,7 +18,7 @@ export const createNode = ({
         language
     }
 }) => {
-    const {propsToSave, mixinsToAdd} = getDataToMutate({}, values, sections, language);
+    const {propsToSave, mixinsToAdd} = getDataToMutate(nodeData, values, sections, language);
```
dgriffon: This is not the way to fix this issue, `nodeData` in create action is the parent node data and is used by `getDataToMutate` to get the  properties to modify. 
We need to handle it in a different way.

dgriffon:
This issue has been handle with another PR

PR #295, opened 2019-10-16
BACKLOG-11378: add starWorkflow 3dots option

FBerthelot:


src/javascript/Edit/index.js
```diff
@@ -34,53 +34,62 @@ export const registerActions = actionsRegistry => {
         target: ['editHeaderActions:1']
     });
 
-    actionsRegistry.add('startWorkflowMainButton', startWorkflowMainButton, {
+    actionsRegistry.add('startWorkflowMainButton', startWorkflow(true), {
```
jkevan: instead of passing true to the component, you can just set a new property:

isMainButton: true
inside the action, like the buttonLabel, buttonIcon, target, etc ...

it will be passed to the action context, and you can reuse it directly in the action code using:

context.isMainButton

PR #327, opened 2019-11-04
BACKLOG-11509 : improve getContentTypesAsTree to use provided node co…

dgriffon:
…nstraints

## JIRA

https://jira.jahia.org/browse/BACKLOG-11509

## Tests

- [x] Integration Tests

## Checklist

- [ ] Performance
- [ ] Migration
- [ ] Code maintainability

## Documentation

- [ ] Inline documentation

### DIFF
--- a/src/main/java/org/jahia/modules/contenteditor/utils/ContentEditorUtils.java
+++ b/src/main/java/org/jahia/modules/contenteditor/utils/ContentEditorUtils.java
@@ -1,6 +1,7 @@
 package org.jahia.modules.contenteditor.utils;
 
 import org.jahia.data.templates.JahiaTemplatesPackage;
+import org.jahia.modules.graphql.provider.dxm.DataFetchingException;
 import org.jahia.registries.ServicesRegistry;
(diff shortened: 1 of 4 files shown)


src/main/java/org/jahia/modules/contenteditor/utils/ContentEditorUtils.java
```diff
@@ -181,16 +239,7 @@ private static boolean checkPermissionForType(ExtendedNodeType type, JCRNodeWrap
...
-        if (nodeTypes != null) {
-            for (String nodeType : nodeTypes) {
-                if (type.isNodeType(nodeType)) {
-                    return true;
-                }
-            }
-            return false;
-        }
-
-        return true;
```
jkevan: You changed the logic of this function, now if List<String> nodeTypes is null it will crash NPE.
Actually the test ContentEditorUtilsTest.getNodetypesTest() is failing because of that

dgriffon: Indeed, thank you, last commit fix the issue

PR #329, opened 2019-11-05
BACKLOG-11513: add linear progress bar when polling publication info

jkevan:


src/javascript/EditPanel/EditPanelContent/EditPanelContent.jsx
```diff
@@ -7,6 +7,7 @@ import {connect} from 'react-redux';
 import {withStyles} from '@material-ui/core';
 import {translate} from 'react-i18next';
 import {PreviewContainer} from './PreviewContainer';
+import PublicationInfoProgress from '../../PublicationInfo/PublicationInfo.progress';
```
dgriffon: Can you use an alias from `PublicationInfo` instead of relative path?

PR #339, opened 2019-11-08
BACKLOG-11433 fix a11y issue with navigate inside choice list items

Taybou:


src/javascript/DesignSystem/MultipleInput/components/Option.jsx
```diff
@@ -1,33 +1,10 @@
...
     return (
-        <MenuItem
-            buttonRef={ref}
-            component="div"
-            className={classes.item}
-            {...innerProps}
-        >
-            {children}
-        </MenuItem>
+        <components.Option {...props} buttonRef={ref}/>
```
FBerthelot: Pour tu n'utilise plus le design systeme ?

Taybou: component doesn't works well with a11y 🗡

PR #349, opened 2019-11-14
BACKLOG-11607 : set license headers

dgriffon:
https://jira.jahia.org/browse/BACKLOG-11607


src/main/java/org/jahia/modules/contenteditor/api/forms/EditorFormFieldValue.java
```diff
@@ -1,3 +1,46 @@
+/*
+ * ==========================================================================================
```
FBerthelot: Are you sure you want to include this at the begining of each file ?

LICENCE.md file at the root of this repository seems more "clean" to me

dgriffon: it's a long time discussion regarding what we have to do on it. 
We end to add it to each file.

PR #357, opened 2019-11-21
BACKLOG-11629 add email validation

Taybou:


src/javascript/Validation/validation.js
```diff
@@ -55,6 +55,20 @@ const dateFieldValidation = (values, field) => {
...
 };
 
+const emailFieldValidation = (values, field) => {
+    const error = 'invalidEmail';
+
+    const fieldValues = Array.isArray(values[field.name]) ? values[field.name] : [values[field.name]];
+    const constraints = field.valueConstraints && field.valueConstraints.map(constraint => constraint.value.string);
+
+    if (constraints && constraints.length > 0 && field.requiredType === 'STRING') {
+        // If one date is invalid, error !
```
jkevan: date ?

src/javascript/Validation/validation.js
```diff
@@ -55,6 +55,20 @@ const dateFieldValidation = (values, field) => {
     return respectEachConstraint ? undefined : error;
 };
 
+const emailFieldValidation = (values, field) => {
+    const error = 'invalidEmail';
+
+    const fieldValues = Array.isArray(values[field.name]) ? values[field.name] : [values[field.name]];
```
jkevan: can be moved inside the next if for optimization.
if we dont have constraint it's not necessary to transform value into array.

Also can we not detect if field.multiple ? instead of Array.isArray(values[field.name]) ?

jkevan:
Also we need to check if error message is correctly displayed from resource bundles.
For exemple JCR prop:

j:email on jnt:user is using RB for error message:

jnt_user.j_email=Email
jnt_user.j_email.constraint.error.message=The input does not conform to valid email syntax

We may need to improve graphQL form API to get this error message.

src/javascript/Validation/validation.spec.js
```diff
@@ -373,6 +373,52 @@ describe('validation', () => {
...
+            expect(validate(sections)(values)).toEqual({
+                field1: undefined,
+                field2: undefined,
+                field3: undefined,
+                field4: undefined
+            });
+        });
+    });
+
     describe('validate order', () => {
```
FBerthelot: You need to add one test about the validation order with mandatory ;)

src/javascript/Validation/validation.spec.js
```diff
@@ -373,6 +373,52 @@ describe('validation', () => {
...
+                requiredType: 'STRING',
+                valueConstraints: [
+                    {value: {string: '^$|[A-Za-z0-9._%+-]+@(?:[A-Za-z0-9-]+\\.)+[A-Za-z]{2,}'}}
+                ]
+            });
+            const values = {
+                field1: 'bonjour@support.fr',
+                field2: 'hallo@support.de',
+                field3: 'hello@support.en',
+                field4: 'world@support.com'
```
FBerthelot: is it valid when field is null, empty string or undefined ?

src/javascript/Validation/validation.spec.js
```diff
@@ -373,6 +373,52 @@ describe('validation', () => {
...
+
+            expect(validate(sections)(values)).toEqual({
+                field1: 'invalidEmail',
+                field2: 'invalidEmail',
+                field3: 'invalidEmail',
+                field4: 'invalidEmail'
+            });
+        });
+
+        it('should not trigger error when is valid email', () => {
```
FBerthelot: Need to add one test for field mutliple

it's change everything as value are in arrays

src/javascript/Validation/validation.spec.js
FBerthelot: Also, need to test for multiple when there is no value in it

src/javascript/Validation/validation.spec.js
```diff
@@ -373,6 +373,122 @@ describe('validation', () => {
         });
     });
 
+    describe('email', () => {
+        it('should trigger error when is not valid email', () => {
```
FBerthelot:
```suggestion
        it('should trigger error when is not valid pattern', () => {
```

Je ne le fais pas pour la suite, t'as compris l'idée :-)

src/javascript/Validation/validation.spec.js
```diff
@@ -373,6 +373,122 @@ describe('validation', () => {
...
+
+            expect(validate(sections)(values)).toEqual({
+                field1: 'invalidEmail',
+                field2: 'invalidEmail',
+                field3: 'invalidEmail',
+                field4: 'invalidEmail'
+            });
+        });
+
+        it('should trigger error when is not valid email and fields are mandatory', () => {
```
FBerthelot: Test a mettre dans le describe "validation order"

src/javascript/Validation/validation.js
```diff
@@ -55,6 +55,21 @@ const dateFieldValidation = (values, field) => {
     return respectEachConstraint ? undefined : error;
 };
 
+const emailFieldValidation = (values, field) => {
+    const error = 'invalidEmail';
```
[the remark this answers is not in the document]
FBerthelot: Ici il manque aussi la traduction pour l'erreur =>  https://github.com/Jahia/content-editor/blob/master/src/main/resources/javascript/locales/fr.json#L98

src/javascript/Validation/validation.js
```diff
@@ -55,6 +55,21 @@ const dateFieldValidation = (values, field) => {
...
+    const error = 'invalidPattern';
+
+    const constraints = field.valueConstraints && field.valueConstraints.map(constraint => constraint.value.string);
+
+    if (constraints && constraints.length > 0 && field.requiredType === 'STRING') {
+        const fieldValues = field.multiple ? values[field.name] : [values[field.name]];
+
+        // If one pattern is invalid, error!
+        if (fieldValues.some(emailValue => !RegExp(constraints[0]).test(String(emailValue).toLowerCase()))) {
+            return error;
```
dgriffon: Why `constraints[0]`? JCR Spec says that the value must match a least one of the constraints. 
Please rename emailValue to something less generic.

Taybou: Ah, you'd like to say, we must validate all the constraints? and should all be regex?

src/javascript/Validation/validation.js
```diff
@@ -55,6 +55,27 @@ const dateFieldValidation = (values, field) => {
...
+const patternFieldValidation = (values, field) => {
+    const error = 'invalidPattern';
+
+    const constraints = field.valueConstraints && field.valueConstraints.map(constraint => constraint.value.string);
+
+    if (constraints && constraints.length > 0 && field.requiredType === 'STRING') {
+        const fieldValues = field.multiple ? values[field.name] : [values[field.name]];
+
+        // If one pattern is invalid, error!
+        if (fieldValues.some(value =>
```
dgriffon: It is not one fieldValue that needs to match the constraint.
it is each field value that needs to match one of the constraint. 
We also need test for this use cases:
- several constraint with no match
- several constraint with only one match
- several values with one that does not match any constraints
- several values with all match of the constraints

PR #359, opened 2019-11-22
Backlog 11673 override error message

Taybou:


Taybou:
I guess it's not possible to test the resource bundle resolution in java unit test.

PR #385, opened 2019-12-12
QA-12407 style card

Taybou:


src/javascript/DesignSystem/Card/Card.jsx
```diff
@@ -23,7 +23,7 @@ const styles = theme => ({
         }
     },
     containerSelected: {
-        border: `1px solid ${theme.palette.brand.alpha}`
+        border: `1px solid ${theme.palette.brand.alpha}!important`
```
FBerthelot: I think we can do a better and stronger CSS selector to avoid `!important` :)

FBerthelot:
```suggestion
        "article&":  {
              border: `1px solid ${theme.palette.brand.alpha}`
         }
```

Taybou: Okay :D

PR #395, opened 2019-12-18
BACKLOG-11935 : fix display of tabs from custom configuration

dgriffon:
https://jira.jahia.org/browse/BACKLOG-11935


jkevan:
no tests on GWT hack for CE too complicated for something we wont support forever

PR #438, opened 2020-01-20
BACKLOG-12112 Fix webpack public path, rename uiLang for consistency …

tdraier:
…with CMM, dx-extend jahia-ui-root

## JIRA

https://jira.jahia.org/browse/PROJECT-XXX

## Tests

- [ ] Integration Tests

## Checklist

- [ ] Performance
- [ ] Migration
- [ ] Code maintainability

## Documentation

- [ ] Inline documentation


package.json
```diff
@@ -30,7 +30,7 @@
     "author": "Jahia Core Team <core@jahia.com>",
     "license": "SEE LICENSE IN LICENSE.txt FILE",
     "dx-extends": {
-        "@jahia/content-manager": "^1.3.0"
+        "@jahia/jahia-ui-root": "^0.0.1"
```
FBerthelot: jahia-ui-root is already deployed ?

src/javascript/Create/Create.container.jsx
```diff
@@ -14,22 +14,22 @@ const mapDispatchToProps = dispatch => ({
 });
 
 const mapStateToProps = state => {
-    const contentEditorUiLang = Constants.supportedLocales.includes(state.uiLang) ?
-        state.uiLang :
+    const contentEditorUiLang = Constants.supportedLocales.includes(state.uilang) ?
```
FBerthelot: Why doing this kind of change ?

src/javascript/useI18n.js
```diff
@@ -2,6 +2,7 @@ import {useState, useEffect} from 'react';
 import {useTranslation} from 'react-i18next';
 
 export const useI18nCENamespace = () => {
+    console.log('Load namespace for CE');
```
FBerthelot: Useless log

src/javascript/register.jsx
```diff
@@ -36,3 +34,5 @@ registry.add('create-route', {
     path: `/:siteKey/:lang/${Constants.routes.baseCreateRoute}`,
     render: () => <ContentEditor mode="create"/>
 });
+
+console.log('%c Content Editor is activated', 'color: #3c8cba');
```
FBerthelot: useless log please keep the .debug log

package.json
tdraier: deployed manually, yes

src/javascript/Create/Create.container.jsx
AKarmanov: Because it needed to be inline with CMM and in CMM it was a mix of uiLang and uilang until I made it all lowercase. We can go whichever way we want I guess (lower or upper), but I believe there needs to be some kind of consistency between projects especially if we share constants like that. Also in jahia-private we have `public static final String UI_LANGUAGE = "uilang";` I happened to be working in all of these projects recently and the inconsistency drove me nuts.

PR #473, opened 2020-02-05
BACKLOG-12293 add field description label

Taybou:


src/javascript/EditPanel/EditPanelContent/FormBuilder/Section/FieldSet/Field/Field.jsx
```diff
@@ -106,6 +106,18 @@ export const FieldCmp = ({classes, inputContext, idInput, selectorType, field, s
...
+                        direction="row"
+                        justify="space-between"
+                        alignItems="center"
+                    >
+                        <Grid item>
+                            <Typography color="beta" variant="delta">
+                                {field.description}
+                            </Typography>
+                        </Grid>
+                    </Grid>
```
dgriffon: Shouldn't we not display the component when there is no description set?

PR #477, opened 2020-02-11
BACKLOG-12335 : fix : preview and back button

dgriffon:
https://jira.jahia.org/browse/BACKLOG-12335


Taybou:
Maybe we should rename the file name, what did you think?

dgriffon:
About the file renaming, we are still using a redux action, we only provide it differently by using the registry

PR #481, opened 2020-02-12
BACKLOG-12304 : adapt component to use tag suggestions query

jsinovassin:
https://jira.jahia.org/browse/BACKLOG-12304


src/javascript/EditPanel/EditPanelContent/FormBuilder/Section/FieldSet/Field/SelectorTypes/Tag/Tag.jsx
```diff
@@ -5,16 +5,38 @@ import React from 'react';
...
+    const suggestTags = async inputValue => {
+        let variables = {
+            prefix: inputValue,
+            limit: '100',
+            startPath: '/sites/' + site
+        };
+
+        const val = await client.query({query: getSuggestionsTagsQuery, variables: variables});
+
+        return val.data.tag.suggest.map(element => {
```
jkevan: missing safety check on val.data undefined.
in case the query do not return data, we should return an empty array

PR #482, opened 2020-02-12
BACKLOG-12363: fix open in new tab on ref field

jkevan:


FBerthelot:
It would be great to have a variable with `/module/moonstone`. This value will change soon.

jkevan:
Even if it change it will still work, if the variable urlbase is correctly set in the DX context, so I thgink it's better to not store /module/moonstone in a variable on CE side.
it's the DX context that have to provide the current context url

PR #486, opened 2020-02-14
BACKLOG-12300: Upgrade react-apollo

FBerthelot:
## JIRA

https://jira.jahia.org/browse/PROJECT-XXX

## Tests

- [ ] Integration Tests

## Checklist

- [ ] Performance
- [ ] Migration
- [ ] Code maintainability

## Documentation

- [ ] Inline documentation


src/javascript/utils/compose.js
```diff
@@ -0,0 +1 @@
+export const compose = (...funcs) => funcs.reduce((a, b) => (...args) => a(b(...args)), arg => arg);
```
Taybou: I suggest to move this func to @jahia/data-helper

FBerthelot: Data-helper should only be composed of gql queries, react-hooks and so on... So now, It's not an option to add compose function into data-helper function.

PR #492, opened 2020-02-19
BACKLOG-12447 fix issue with Category.jsx

Taybou:


src/javascript/EditPanel/EditPanelContent/FormBuilder/Section/FieldSet/Field/SelectorTypes/Category/Category.jsx
```diff
@@ -46,7 +46,7 @@ const Category = ({field, value, id, editorContext}) => {
     });
 
     const tree = useMemo(() => {
-        if (!data.jcr) {
+        if (error || loading || !data.jcr) {
```
dgriffon: We use to display a ProgressOverlay when loading .. why isn't it the case here?
```
  if (loading ||!data.jcr) {
        return <ProgressOverlay/>;
    }
```

Taybou: It's bellow, and this is only to fix the issue with the meme...

I don't know why "MEMO" it's used here.

PR #495, opened 2020-02-20
BACKLOG-12407 : make edit button from page composer to display conten…

dgriffon:
https://jira.jahia.org/browse/BACKLOG-12407

### DIFF
--- a/src/javascript/register.jsx
+++ b/src/javascript/register.jsx
@@ -44,4 +44,11 @@ registry.add('callback', 'content-editor', {
     }
 });
 
+// Register GWT Hooks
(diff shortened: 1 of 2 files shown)


src/javascript/register.jsx
```diff
@@ -44,4 +44,11 @@ registry.add('callback', 'content-editor', {
     }
 });
 
+// Register GWT Hooks
+window.top.jahiaGwtHook = {
```
Taybou: where will this function be called? have you completed the implementation?

dgriffon: this function is called by GWT

PR #497, opened 2020-02-21
BACKLOG-12455 : use picker component to show children nodes

jsinovassin:
## JIRA

https://jira.jahia.org/browse/BACKLOG-12455


src/main/resources/javascript/locales/en.json
```diff
@@ -158,7 +158,8 @@
       },
       "section":{
         "listAndOrdering": {
-          "title": "Content list & ordering"
```
jkevan: missing other translations

PR #507, opened 2020-02-27
BACKLOG-12514 : expose Picker as an API

dgriffon:
https://jira.jahia.org/browse/BACKLOG-12514


dgriffon:
> Move pickeer API into richText cmp

Done!

PR #514, opened 2020-02-28
BACKLOG-12537 : add the modal to allows to copy values from a language

jsinovassin:
## JIRA

https://jira.jahia.org/browse/BACKLOG-12537

### DIFF
--- a/src/javascript/Edit/copyLanguage/copyLanguage.utils.js
+++ b/src/javascript/Edit/copyLanguage/copyLanguage.utils.js
@@ -1,0 +1,12 @@
+
+/**
+ * This function get the full name of a language by his id
+ *
+ * @param languages list of available languages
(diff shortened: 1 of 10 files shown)


src/javascript/Edit/copyLanguage/copyLanguage.utils.js
```diff
@@ -0,0 +1,12 @@
+
+/**
+ * This function get the full name of a language by his id
+ *
+ * @param languages list of available languages
+ * @param id language id to get
+ */
+export function getFullLanguageName(languages, id) {
```
dgriffon: please add test to that utility class, thank you

jsinovassin: Tests added

PR #534, opened 2020-03-09
Backlog 12681 title

Taybou:


src/javascript/EditPanel/EditPanel.jsx
```diff
@@ -113,7 +113,7 @@ const EditPanelCmp = ({formik, title, classes, notificationContext, client}) =>
                                     return <Button context={context}/>;
                                 }}
                             />
-                            {title}
+                            {title.slice(0, 100) + '...'}
```
FBerthelot: What happen if there is less than 100char to the title ?

I think you can do unit test about that + handle the case when title is less than 100char

Taybou: "florent".slice(0,100) => "florent"

And it will be updated in the "header style" story with much better handling ...

FBerthelot: `{title.slice(0, 100) + '...'}` if title is florent, then it will write `florent...`, not `florent`. It's strange isn't it ?

PR #545, opened 2020-03-11
BACKLOG-12743 style CE header

Taybou:


src/javascript/DesignSystem/ContentLayout/ContentHeader/ContentHeader.test.jsx
```diff
@@ -3,51 +3,8 @@ import {shallow} from '@jahia/test-framework';
 import ContentHeader from './ContentHeader';
 
 describe('Content header', () => {
-    it('Should render with everything', async () => {
```
FBerthelot: Nope, don't delete tests

Taybou: I update the component, so I must delete and or update the test :D

FBerthelot: Yes please update the test to the new behaviour.

All test that you delete, you can move them to another place where the unchanged test belong.

Taybou: the is no other places, the code doesn't exist anymore

PR #578, opened 2020-03-26
BACKLOG-12921 : add data sel role for selenium

jsinovassin:
## JIRA

https://jira.jahia.org/browse/BACKLOG-12921

### DIFF
--- a/src/javascript/Edit/Edit.actions.js
+++ b/src/javascript/Edit/Edit.actions.js
@@ -52,7 +52,7 @@ export const registerEditActions = actionsRegistry => {
         menuTarget: 'ContentEditorHeaderActions'
     });
 
-    actionsRegistry.add('action', 'startWorkflow3dots', startWorkflow, {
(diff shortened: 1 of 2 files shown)


src/javascript/Edit/Edit.actions.js
```diff
@@ -52,7 +52,7 @@ export const registerEditActions = actionsRegistry => {
         menuTarget: 'ContentEditorHeaderActions'
     });
 
-    actionsRegistry.add('action', 'startWorkflow3dots', startWorkflow, {
+    actionsRegistry.add('action', 'startWorkflowHeader', startWorkflow, {
```
Taybou: why updating this name?

jsinovassin: because it's not a 3 dots menu anymore, the start worklow action is still in the header

Taybou: all right!

PR #582, opened 2020-03-27
DMF-3778 add content editor event handler

Taybou:
## JIRA

https://jira.jahia.org/browse/DMF-3778

## Description
Add content editor event handler

## Tests

- [ ] Integration Tests

## Checklist

- [x] Security (in particular for changes to authentication, authorization, data fetching, ...)
- [x] Performance
- [ ] Migration
- [x] Code maintainability

## Documentation

- [ ] Inline documentation


src/javascript/Api/ContentEditor.api.jsx
```diff
@@ -122,6 +122,11 @@ const ContentEditorApiCmp = ({classes, client}) => {
                 window.top.authoringApi.refreshContent();
             }
 
+            // Refresh contentEditorEventHandler
+            if (window.top.contentEditorEventHandler) {
```
jkevan: it could be better to use native js event here, probably some other apps will require this kind of info.
if we use native js events you could send an event here, it will be catch by any declared listeners.

```
// Listen to get the event: (From JExperience)
elem.addEventListener('build', function (e) { ... }, false);

// Dispatch the event: (From Content Editor)
var event = new Event('build');
elem.dispatchEvent(event);
```
This way if at  any time, somebody else than jexperience need to be aware of CE update/create, it will juste have to register a new event listener.
We just need to find the good element to use to listen and dispatch events.
I think the body of document could be a good choice.

PR #583, opened 2020-03-27
Backlog 12856 manage option wip on create

jsinovassin:
## JIRA

https://jira.jahia.org/browse/BACKLOG-12856


dgriffon:
This will be done using a WIP service, sorry but we have to close that pull request
see https://github.com/Jahia/content-editor/pull/584

PR #591, opened 2020-03-30
BACKLOG-12967: system name encoding JCR reserverd characters

jkevan:
system name encoding for reserved JCR characters
covered by react tests

### DIFF
--- a/src/javascript/utils/helper.js
+++ b/src/javascript/utils/helper.js
@@ -5,3 +5,45 @@ export const truncate = (string, num) => {
 
     return string.slice(0, num) + '...';
 };
+
+/**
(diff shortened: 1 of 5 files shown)


src/javascript/utils/helper.js
```diff
@@ -5,3 +5,45 @@ export const truncate = (string, num) => {
 
     return string.slice(0, num) + '...';
 };
+
+/**
+ * JCR system name doesnt allow this characters: ['/', ':', '[', ']', '|', '*']
```
jsinovassin: Do we accept the '\' character?

dgriffon: yes, invalid character are from that list:
https://docs.adobe.com/docs/en/spec/jcr/2.0/3_Repository_Model.html#3.2.2%20Local%20Names

PR #598, opened 2020-03-31
BACKLOG-13035 : add unsaved info chip when a content is modified

jsinovassin:
## JIRA

https://jira.jahia.org/browse/BACKLOG-13035


src/javascript/EditPanel/header/UpperSection.jsx
```diff
@@ -131,6 +131,11 @@ export const HeaderUpperSection = ({title, actionContext}) => {
                     <PublicationInfoBadge/>
                     <LockInfoBadge/>
                     <WipInfoChip wipInfo={wipInfo}/>
+                    {(formik.dirty || mode === 'create') &&
```
dgriffon, line 134: use mode constant

PR #611, opened 2020-04-03
BACKLOG-13076 : hide advanced option tab on create

jsinovassin:
## JIRA

https://jira.jahia.org/browse/BACKLOG-13076

## Tests


src/javascript/EditPanel/tabbar/TabBar.utils.js
```diff
@@ -0,0 +1,10 @@
+import {Constants} from '~/ContentEditor.constants';
+
+export const hasToRenderAction = (actionName, mode) => {
+    const actionCondition = {
+        edit: true,
+        advanced: mode === Constants.routes.baseEditRoute
+    };
+
+    return actionCondition[actionName];
+};
```
jkevan, line 10: It's really hard to understand what this code is doing

jkevan: It would be more easy to store the info, if it's displayable or not in edit and/or create inside the action definition inside file: EditPanel.action.js instead of hardcoding stuff here, this is configuration, it should be part of the action definition and store inside the context, then checked by the tabbar.action directly.

PR #622, opened 2020-04-08
BACKLOG-13088: fix pastille postion

FBerthelot:
## JIRA

https://jira.jahia.org/browse/BACKLOG-13088

## Description

Fix pastille position and split save-action from publish actions

### DIFF
--- a/src/javascript/EditPanel/header/UpperSection.jsx
+++ b/src/javascript/EditPanel/header/UpperSection.jsx
@@ -1,9 +1,9 @@
 import React from 'react';
 import PropTypes from 'prop-types';
 
-import {Button, Typography, ButtonGroup, Chip} from '@jahia/moonstone';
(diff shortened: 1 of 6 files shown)


src/javascript/EditPanel/header/UpperSection.jsx
```diff
@@ -76,9 +59,25 @@ export const HeaderUpperSection = ({title, actionContext}) => {
                 </div>
 
                 <div className={styles.headerRight}>
+                    <div className={styles.saveActions}>
+                        <DisplayActions context={{
```
dgriffon, line 63: Can we have several actions here ? Can't we use DisplayAction (without "s") instead

FBerthelot: Yes, we can have multiple actions here event if it's not a button group. Romain wanted that.

PR #649, opened 2020-04-21
BACKLOG-13169 : remove wip info when modifying a site node

jsinovassin:
## JIRA

https://jira.jahia.org/browse/BACKLOG-13169


src/javascript/Edit/save/save.gql-mutation.js
```diff
@@ -24,7 +25,7 @@ export const SavePropertiesMutation = gql`
                     delete(language: $language)
                 }
                 removeMixins(mixins: $mixinsToDelete)
-                mutateWipInfo(wipInfo:$wipInfo)
+                mutateWipInfo(wipInfo:$wipInfo) @include(if: $shouldSetWip)
```
dgriffon, line 28: We should not only prevent to save it but also not display the button to set the WIP on the content

PR #657, opened 2020-04-22
BACKLOG-13082 : remove / character in the system name and replace *

jsinovassin:
## JIRA

https://jira.jahia.org/browse/BACKLOG-13082
## Tests

### DIFF
--- a/src/javascript/utils/helper.js
+++ b/src/javascript/utils/helper.js
@@ -14,16 +14,17 @@ export const truncate = (string, num) => {
  * @returns {*} the encoded system name
  */
 export const encodeSystemName = systemName => {
+    systemName = systemName.replace('/', '');
     if (systemName && systemName.length > 0) {
-        // First Step: Encode characters using encodeURI js API
         let encodedSystemName = '';
+        // First Step: Encode characters using encodeURI js API
         for (let i = 0; i < systemName.length; i++) {
             const charToEncode = systemName.charAt(i);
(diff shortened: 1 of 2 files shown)


src/javascript/utils/helper.js
```diff
@@ -14,16 +14,17 @@ export const truncate = (string, num) => {
...
  */
 export const encodeSystemName = systemName => {
+    systemName = systemName.replace('/', '');
     if (systemName && systemName.length > 0) {
-        // First Step: Encode characters using encodeURI js API
         let encodedSystemName = '';
+        // First Step: Encode characters using encodeURI js API
         for (let i = 0; i < systemName.length; i++) {
             const charToEncode = systemName.charAt(i);
             encodedSystemName += ['%', '/', ':', '[', ']', '|'].includes(charToEncode) ? encodeURIComponent(charToEncode) : charToEncode;
```
Taybou, line 23: no need for "/" anymore

jsinovassin: The "/" character in the system name causes a problem because the system name is used in the URL, and when we access to a content with a "/" it's like we try to get a subnode instead of the current node
 In the edit engine this character was removed too.

Taybou: @jsinovassin Yes, I understand, but this check of ['%', '/', ':', '[', ']', '|'].includes(charToEncode), the charToEncode will never be '/', because you have remove it before in your systemName variable.

PR #666, opened 2020-04-23
BACKLOG-13264 : register / unregister listeners that trigger modal di…

dgriffon:
https://jira.jahia.org/browse/BACKLOG-13264
Clean up listener registration that handle modal window for pending changes

### DIFF
--- a/src/javascript/EditPanel/EditPanel.jsx
+++ b/src/javascript/EditPanel/EditPanel.jsx
@@ -25,30 +25,40 @@ const EditPanelCmp = ({formik, title, notificationContext, client}) => {
         }
 
         const handleBeforeUnloadEvent = ev => {
-            if (formik.dirty) {
-                ev.preventDefault();
-                ev.returnValue = '';
+            ev.preventDefault();
+            ev.returnValue = '';
+        };
+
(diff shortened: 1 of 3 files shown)


src/javascript/EditPanel/EditPanel.jsx
```diff
@@ -31,24 +31,36 @@ const EditPanelCmp = ({formik, title, notificationContext, client}) => {
...
-        return () => {
+        const unregisterListeners = () => {
             window.removeEventListener('beforeunload', handleBeforeUnloadEvent);
+            if (envProps.unregisterListeners) {
+                envProps.unregisterListeners();
+            }
         };
+
+        if (formik.dirty) {
+            registerListeners();
```
jkevan, line 50: there is a double check to formik.dirty ? one here and one inside ContentEditor.redux.jsx (registerListeners)
May be we can remove the check from ContentEditor.redux.jsx (registerListeners) ?
and may be we do not need formik anymore inside ContentEditor.redux.jsx (registerListeners) ?

dgriffon: Yes, the check was not useful. 
Thank you!

PR #677, opened 2020-04-27
QA-12730 : display non selectable content in pickers

dgriffon:
https://jira.jahia.org/browse/QA-12730
This allows to display non selectable items in pickers. Link picker use this configuration. 
Covered with unit test.

### DIFF
--- a/src/javascript/EditPanel/EditPanelContent/FormBuilder/Section/FieldSet/Field/SelectorTypes/Picker/PickerDialog/Views/List/List.spec.jsx
+++ b/src/javascript/EditPanel/EditPanelContent/FormBuilder/Section/FieldSet/Field/SelectorTypes/Picker/PickerDialog/Views/List/List.spec.jsx
@@ -170,8 +170,8 @@ describe('PickerDialog - List view', () => {
         )
             .find('ContentTable');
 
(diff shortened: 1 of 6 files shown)


src/javascript/EditPanel/EditPanelContent/FormBuilder/Section/FieldSet/Field/SelectorTypes/Picker/PickerDialog/Views/List/List.spec.jsx
```diff
@@ -152,6 +155,25 @@ describe('PickerDialog - List view', () => {
...
+        defaultProps.pickerConfig.showOnlyNodesWithTemplates = true;
+
+        const cmp = shallowWithTheme(
+            <List {...defaultProps}/>,
+            {},
+            dsGenericTheme
+        )
+            .find('ContentTable');
+
+        expect(cmp.props().data[0].selectable).toBeFalsy();
```
jsinovassin, line 173: Can not be null so you can use .toBe(false) and toBe(true)

dgriffon: Thank you for the hint!

PR #679, opened 2020-04-28
BACKLOG-13383 : flush cache on save

jsinovassin:
## JIRA

https://jira.jahia.org/browse/BACKLOG-13383


src/javascript/Edit/save/save.action.jsx
```diff
@@ -27,6 +29,8 @@ const Save = ({context, render: Render, loading: Loading}) => {
                         return formik
                             .submitForm()
                             .then(() => {
+                                // TODO BACKLOG-13406 avoid refretch if possible
+                                refetchFormData();
```
dgriffon, line 33: I am a bit worried about that refetch that will trigger a redraw of Content editor, additional to the cache flushed below, can you ensure that there is only one drawing of Content Editor after a save? See BACKLOG-13364 to ensure that.

jsinovassin: I tried, it's working as before the modification, content editor is rendered once and the lock seems to working as expected.

PR #689, opened 2020-05-07
BACKLOG-13618: fix template set CKEditor custom config

jkevan:


src/javascript/EditPanel/EditPanelContent/FormBuilder/Section/FieldSet/Field/SelectorTypes/RichText/RichText.jsx
```diff
@@ -64,7 +77,7 @@ export const RichTextCmp = ({field, id, value}) => {
     };
 
     const config = {
-        customConfig: ckeditorCustomConfig.replace('$context', window.contextJsParameters.contextPath),
+        customConfig: ckeditorCustomConfig,
```
Taybou, line 80: I think we can keep this here and remove it from above, as the ckeditorCustomConfig will never be undefined or null, al least will be empty string ''

PR #693, opened 2020-05-14
BACKLOG-13690: refetchObservableQueries from client on save, and publ…

jkevan:


src/javascript/PublicationInfo/PublicationInfo.jsx
```diff
@@ -19,6 +19,11 @@ export const usePublicationInfo = (queryParams, t) => {
             if (jobsData && jobsData.endedJobs && jobsData.endedJobs.length > 0) {
                 refetch();
             }
+
+            // Refetch publication info in case of unpublication
+            if (jobsData.type === 'contentUnpublished') {
```
dgriffon, line 24: Is it possible to merge the two `if` to avoid 2 refreshes

PR #695, opened 2020-05-19
BACKGLOG-13711 : handle value constraints correctly when overriding w…

jsinovassin:
…ith json

## JIRA

https://jira.jahia.org/browse/BACKLOG-13711

## Tests

### DIFF
--- a/src/main/java/org/jahia/modules/contenteditor/api/forms/EditorFormFieldValue.java
+++ b/src/main/java/org/jahia/modules/contenteditor/api/forms/EditorFormFieldValue.java
@@ -137,6 +137,11 @@ public class EditorFormFieldValue {
         return type;
     }
 
+    public String getValue() {
+        // Allows ObjectMapper to correctly fill the field when customizing constraint value with json file
+        return value;
+    }
+
     @Override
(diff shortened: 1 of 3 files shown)


src/main/java/org/jahia/modules/contenteditor/api/forms/EditorFormFieldValue.java
```diff
@@ -137,6 +137,11 @@ public String getType() {
         return type;
     }
 
+    public String getValue() {
```
Taybou, line 140: What's the difference between this method and `getStringValue`?

jsinovassin: Those two methods get the same result but ObjectMapper needs (by default) a getter which have the same name of the properties of the class that you want to fill. Before my modification we had a property named `value` with a getter named `getStringValue` so ObjectMapper can not do the mapping automatically.

Taybou: Good to know, thank you :D

PR #712, opened 2020-06-12
Add mixin method to the registry

jsinovassin:
## JIRA

https://jira.jahia.org/browse/BACKLOG-13835

## Tests

- [ ] Integration Tests

## Checklist

- [ ] Performance
- [ ] Migration
- [ ] Code maintainability

## Documentation

- [ ] Inline documentation


src/javascript/ContentEditor.register.Utils.js
```diff
@@ -0,0 +1,65 @@
+export const removeMixinFromSection = (mixin, sections) => {
+    let removedFields = [];
```
dgriffon, line 2: This function needs to be split in two, one to enable / disable a fieldset, the other to move a field to a new location (fieldset / field)

src/javascript/ContentEditor.register.Utils.js
```diff
@@ -0,0 +1,65 @@
...
+
+    sections = sections.map(section => {
+        return {...section,
+            fieldSets: section.fieldSets.map(fieldSet => {
+                if (fieldSet.name === mixin) {
+                    fieldSet.dynamic = true;
+                    fieldSet.activated = true;
+                }
+
+                let updatedFields = fieldSet.fields;
```
dgriffon, line 49: Use the move field function to do this operation

src/javascript/ContentEditor.register.jsx
```diff
@@ -5,6 +5,7 @@ import {Constants} from '~/ContentEditor.constants';
 import ContentEditorApi from '~/Api/ContentEditor.api';
 import ContentEditorRedux from './ContentEditor.redux';
 import {ContentEditorHistoryContextProvider} from '~/ContentEditorHistory/ContentEditorHistory.context';
+import {removeMixinFromSection, addMixinToSection} from '~/ContentEditor.register.Utils';
```
dgriffon, line 8: As we want customers to use the utility function, can we pass the utility functions as an onChange parameter?

src/javascript/ContentEditor.context.js
```diff
@@ -53,6 +63,7 @@ export const withContentEditorDataContextProvider = (formQuery, formDataAdapter)
             return <ProgressOverlay/>;
         }
 
+        sections = sections.length === 0 ? formSections : sections;
```
dgriffon, line 66: The setSection has been added because the context was refreshed but should not. 
https://jira.jahia.org/browse/BACKLOG-13837 should work on that

PR #732, opened 2020-06-26
QA-13003 add support for encryptedValue in content editor

Taybou:
## JIRA

https://jira.jahia.org/browse/QA-13003

## Tests

- [x] Unit Tests (Most changes _should_ have unit tests)
- [ ] Integration Tests

## Checklist

- [ ] Performance
- [ ] Migration
- [ ] Code maintainability

## Documentation

- [ ] Inline documentation


src/javascript/SelectorTypes/SelectorTypes.js
```diff
@@ -19,7 +19,15 @@ const adaptDateProperty = (field, property) => {
...
-    ceRegistry.add('selectorType', 'Text', {cmp: Text, supportMultiple: false});
+    ceRegistry.add('selectorType', 'Text', {
+        cmp: Text,
+        supportMultiple: false,
+        adaptValue: (field, property) => {
+            if (field.selectorOptions?.find(option => option.name === 'password')) {
+                return property.encryptedValue;
+            }
+        }
+    });
```
jkevan, line 30: Your Backend API do not provide multiple support for this property to be set, so how can it work here in case the field is multiple ? it will not.
Backend should provide a multiple support.

src/javascript/EditPanel/EditPanel.utils.js
```diff
@@ -134,6 +134,10 @@ export function getDataToMutate({nodeData, formValues, sections, lang}) {
  * @returns {string} the name of the value property to use
  */
 export function getValuePropName(field) {
+    if (!field.multiple && field.selectorOptions?.find(option => option.name === 'password')) {
+        return 'encryptedValue';
+    }
+
```
jkevan, line 140: Please consider review the API, if we use a boolean it's not necessary to do that, instead we have to add a the boolean to encrypt the value in the mutation

PR #743, opened 2020-07-03
BACKLOG-14033: handle jcr:nodeTypeManagememt for mixin(fieldSet) enab…

jkevan:


src/main/java/org/jahia/modules/contenteditor/api/forms/EditorFormField.java
```diff
@@ -396,7 +396,7 @@ public String toString() {
             '}';
     }
 
-    private Boolean mergeBooleanKeepTrue(Boolean value1, Boolean value2) {
+    private static Boolean mergeBooleanKeepTrue(Boolean value1, Boolean value2) {
```
Taybou, line 399: why add static here?

jkevan: I wanted to use it for the new readOnly prop on fieldset for merge, but it was not making sense to allow overriding this prop. So at the end, I didn't use it, but forgot to remove the static.
I think it's ok to keep it static, actually this fcts should be statics.

PR #747, opened 2020-07-06
BACKLOG-13942 - BACKLOG-14006 : rework onChange for init / unMount to…

dgriffon:
https://jira.jahia.org/browse/BACKLOG-13942
https://jira.jahia.org/browse/BACKLOG-14006

Fix cascading onChange
covered by unit tests

### DIFF
--- a/src/javascript/EditPanel/EditPanelContent/FormBuilder/Section/FieldSet/Field/MultipleField.jsx
+++ b/src/javascript/EditPanel/EditPanelContent/FormBuilder/Section/FieldSet/Field/MultipleField.jsx
@@ -25,15 +25,15 @@ const styles = theme => {
 export const MultipleFieldCmp = ({classes, editorContext, inputContext, field, onChange, formik: {values}}) => {
     const {t} = useTranslation();
     const [isInit, setInit] = useState(false);
-    const currentValue = useRef(undefined);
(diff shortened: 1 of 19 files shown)


src/javascript/EditPanel/EditPanelContent/FormBuilder/Section/FieldSet/Field/MultipleField.jsx
```diff
@@ -22,48 +22,41 @@ const styles = theme => {
...
-        default:
-            throw new Error();
-    }
-};
-
-export const MultipleFieldCmp = ({classes, inputContext, field, onChange, formik: {values}}) => {
+export const MultipleFieldCmp = ({classes, editorContext, inputContext, field, onChange, formik: {values}}) => {
     const {t} = useTranslation();
     const [isInit, setInit] = useState(false);
+    const currentValue = useRef(undefined);
```
Taybou, line 28: I cannot understand why using this manipulation? I think we added complexity to code and it will cause some rendering and performance issues.

dgriffon: We are using a ref to be able to read the value from closured function call.

PR #769, opened 2020-07-22
BACKLOG-14062: Moonstone Colours, Language Switcher, Buttons

vindhya:
## JIRA

https://jira.jahia.org/browse/BACKLOG-14062

## Description

* Change darker blue accent colour to Moonstone lighter blue accent colour for many components
* Changed Language Switcher in the header to Moonstone dropdown component
* More buttons changed to Moonstone buttons

## Tests

- [x] Unit Tests (Most changes _should_ have unit tests)

### DIFF
--- a/src/main/resources/javascript/locales/en.json
+++ b/src/main/resources/javascript/locales/en.json
@@ -190,7 +190,7 @@
         "createNewContent": {
           "menu": "New content",
           "contentOfType": "New {{typeName}}",
-          "labelModal": "SELECT CONTENT TYPE",
+          "labelModal": "Select Content Type",
           "btnDiscard": "CANCEL",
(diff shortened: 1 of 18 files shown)


src/main/resources/javascript/locales/en.json
```diff
@@ -190,7 +190,7 @@
         "createNewContent": {
           "menu": "New content",
           "contentOfType": "New {{typeName}}",
-          "labelModal": "SELECT CONTENT TYPE",
+          "labelModal": "Select Content Type",
```
dgriffon, line 193: We need to change the labels in other languages as well

vindhya: Good point! I've now updated French and German as well.

PR #770, opened 2020-07-22
BACKLOG-14188 manage mode when access CE from url

Taybou:
## JIRA

https://jira.jahia.org/browse/BACKLOG-14188

## Tests

- [ ] Integration Tests

## Checklist

- [ ] Performance
- [ ] Migration
- [ ] Code maintainability

## Documentation

- [ ] Inline documentation

### DIFF
--- a/src/javascript/EditPanel/header/ContentBreadcrumb/ContentPath/ContentPath.container.jsx
+++ b/src/javascript/EditPanel/header/ContentBreadcrumb/ContentPath/ContentPath.container.jsx
@@ -27,9 +27,9 @@ const ContentPathContainer = ({path}) => {
     const handleNavigation = path => {
         let newMode = 'pages';
 
-        if (path.startsWith(`/sites/${site}/files/`)) {
+        if (path.startsWith(`/sites/${site}/files`) || path === `/sites/${site}/files`) {
             newMode = 'media';
-        } else if (path.startsWith(`/sites/${site}/contents/`)) {
+        } else if (path.startsWith(`/sites/${site}/contents/`) || path === `/sites/${site}/contents`) {
             newMode = 'content-folders';
         }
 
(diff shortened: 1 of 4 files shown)


src/javascript/EditPanel/header/ContentBreadcrumb/ContentPath/ContentPath.container.jsx
```diff
@@ -23,10 +25,22 @@ const ContentPathContainer = ({path}) => {
     });
 
     const handleNavigation = path => {
+        let newMode = 'content-folders';
+
+        if (path.indexOf('/files') !== -1) {
+            newMode = 'media';
```
dgriffon, line 31: This test is not enough as a content named `filesName` will match files even not being under `/files`.
You need to do the lookup in the right place in the path `/sites/sitename/${type}/`
and not forget about the trailing `/` 

Also the default is pages, as we know that contents are under `contents` and we can have other pages at the same level as `home`

src/javascript/EditPanel/header/ContentBreadcrumb/ContentPath/ContentPath.container.jsx
```diff
@@ -23,10 +25,20 @@ const ContentPathContainer = ({path}) => {
     });
 
     const handleNavigation = path => {
+        let newMode = 'pages';
+
+        if (path.startsWith(`/sites/${site}/files`)) {
```
dgriffon, line 30: you miss the trailing `/` .. if a page is named `filesPage` at root level, it will be detected as a media

src/javascript/EditPanel/header/ContentBreadcrumb/ContentPath/ContentPath.container.jsx
dgriffon: All the cases describe above are not fulfilled. With the current code a path like `/sites/siteKey/files` will be resolved as `pages` instead of `media`

PR #788, opened 2020-07-31
QA-13077, QA-13079: handle server constraint violation error after save

jkevan:


src/javascript/Validation/validation.utils.js
```diff
@@ -35,38 +35,44 @@ export const onServerError = (error, formikActions, notificationContext, t, defa
...
+                    }
                 }
             }
         }
     }
 
-    // No error handling, print error in console and use default error message
-    console.error(error);
-    notificationContext.notify(t(defaultErrorMessage), ['closeButton']);
+    if (notificationErrorMessage) {
```
Taybou, line 75: we will have always default error message, no?

src/main/resources/javascript/locales/fr.json
```diff
@@ -156,7 +156,8 @@
           "required": "",
           "alreadyExist": "Il y a déjà un objet avec ce nom système, veuillez en choisir un autre",
           "maxLength": "{{selectorOptions.maxLength}} caractères maximum",
-          "invalidLink": "Le lien {{0}} est rompu"
+          "invalidLink": "Le lien {{0}} est rompu",
```
Taybou, line 159: c'est correct le mot: rompu?

src/javascript/Validation/validation.utils.js
jkevan: not in case of javax.jcr.ItemExistsException
the error is: content-editor:label.contentEditor.error.changeSystemName

src/javascript/Validation/validation.utils.js
```diff
@@ -35,38 +35,44 @@ export const onServerError = (error, formikActions, notificationContext, t, defa
...
+                graphQLError.extensions && graphQLError.extensions.constraintViolations &&
+                graphQLError.extensions.constraintViolations.length > 0) {
+                // Constraint violation errors
 
-                const extractedErrorData = graphQLError.message.match(/^org.jahia.services.content.PropertyConstraintViolationException: (?<prop>(.*?)): Invalid link(?<link>(.*?))$/);
-                if (extractedErrorData && extractedErrorData.groups && extractedErrorData.groups.prop && extractedErrorData.groups.link) {
-                    const propNameIndex = extractedErrorData.groups.prop.lastIndexOf(' ');
-                    const propName = extractedErrorData.groups.prop.substring(propNameIndex + 1);
+                for (const constraintViolation of graphQLError.extensions.constraintViolations) {
+                    console.log('Constraint violation error: ', constraintViolation);
```
Taybou, line 57: is it necessary, maybe we can convert it to .error?

src/javascript/Validation/validation.utils.js
jkevan: and in case of constraints violations:
'content-editor:label.contentEditor.error.constraintViolations'

src/main/resources/javascript/locales/fr.json
jkevan: Un lien mort, lien brisé, lien cassé ou lien rompu est un lien hypertexte qui redirige vers un site web ou une page web n'existant plus.
source: https://fr.wikipedia.org/wiki/Lien_mort

src/javascript/Validation/validation.utils.js
Taybou: I'm talking about this one:

let notificationErrorMessage = t(defaultErrorMessage);

src/javascript/Validation/validation.utils.js
jkevan: CE do not support NodeConstraintViolation that are not attached to a prop.
That's why I prefer to log them to give details.
.error is fine but not mandatory, as it's mostly logged for informations but nothing is broken in the application, 
I prefer to use .error when and real error happen client side.

PR #790, opened 2020-07-31
BAKLOG-13207 : update jcontent path when system name changed

jsinovassin:
## JIRA

https://jira.jahia.org/browse/BACKLOG-13207

### DIFF
--- a/src/javascript/ContentEditor.redux.jsx
+++ b/src/javascript/ContentEditor.redux.jsx
@@ -1,5 +1,5 @@
 import React from 'react';
-import {connect} from 'react-redux';
+import {connect, useDispatch, useSelector} from 'react-redux';
 import {ContentEditor} from './ContentEditor';
 import PropTypes from 'prop-types';
 import {compose} from '~/utils';
@@ -7,6 +7,9 @@ import {Constants} from '~/ContentEditor.constants';
 import {useContentEditorHistory} from '~/ContentEditorHistory';
 import {useTranslation} from 'react-i18next';
 import {withApollo} from 'react-apollo';
+import {replaceOpenedPath} from '~/JContent.redux-actions';
+import {useContentEditorHistoryContext} from '~/ContentEditorHistory/ContentEditorHistory.context';
+import {registry} from '@jahia/ui-extender';
 
 const mapStateToProps = state => {
     return {
@@ -17,18 +20,42 @@ const mapStateToProps = state => {
 
 const ContentEditorReduxCmp = ({client, mode, uuid, lang, uilang, site, contentType}) => {
     const {redirect, hasHistory, exit, registerBlockListener, unRegisterBlockListener} = useContentEditorHistory();
+    const {storedLocation, setStoredLocation} = useContentEditorHistoryContext();
+    const {openPaths} = useSelector(state => ({openPaths: state.jcontent.openPaths}));
+    const dispatch = useDispatch();
     const {t} = useTranslation();
     // Sync GWT language
     if (window.top.authoringApi.switchLanguage) {
         window.top.authoringApi.switchLanguage(lang);
     }
 
+    const rename = (node, mutateNode) => {
+        if (mutateNode && mutateNode.rename) {
+            if (storedLocation && storedLocation.location) {
+                if (openPaths) {
+                    dispatch(replaceOpenedPath(openPaths.map(openPath => openPath.replace(node.path, mutateNode.rename))));
+                }
+
+                const {site: currentSite, language, mode: currentMode, path} = registry.get('jcontent', 'utils').extractParamsFromUrl(storedLocation.location.pathname, storedLocation.location.search);
+                const newStoredLocation = {
+                    ...storedLocation,
+                    location: {
+                        ...storedLocation.location,
+                        pathname: registry.get('jcontent', 'utils').buildUrl(currentSite, language, currentMode, path.replace(node.path, mutateNode.rename))
+                    }
+                };
+                setStoredLocation(newStoredLocation);
+                return newStoredLocation;
+            }
+        }
+    };
+
     const envProps = {
         setUrl: (mode, language, uuid, contentType) => redirect({mode, language, uuid, rest: contentType}),
-        back: () => {
+        back: (uuid, operator, node, mutateNode) => {
             client.cache.flushNodeEntryById(uuid);
-
-            exit();
+            const overridedStoredLocation = rename(node, mutateNode);
+            exit(overridedStoredLocation);
         },
         disabledBack: () => !hasHistory(),
         setLanguage: language => redirect({language}),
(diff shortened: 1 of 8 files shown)


src/javascript/Edit/Edit.jsx
```diff
@@ -39,13 +46,31 @@ export const EditCmp = ({
                 sections,
                 values
             },
-            editCallback: nodeUuid => {
+            editCallback: (node, mutateNode) => {
```
jkevan, line 49: Rename handling should be done before the editCallback (because the edit callback is handling the close, so it will be based on history and we need to update the history before that)

src/javascript/Edit/Edit.jsx
```diff
@@ -39,13 +46,31 @@ export const EditCmp = ({
...
+                        envEditCallback(node.uuid, contentEditorConfigContext);
+                    }
+                }
+
+                if (mutateNode.rename) {
+                    if (storedLocation && storedLocation.location) {
+                        const newNameStartPosition = mutateNode.rename.lastIndexOf('/') + 1;
+                        const newName = mutateNode.rename.substring(newNameStartPosition, mutateNode.rename.length);
+                        if (openPaths) {
+                            dispatch(replaceOpenedPath(openPaths.map(openPath => openPath.replace(node.name, newName))));
```
jkevan, line 64: Dangerous, do not rename other element of the path that should not be rename,
Replace the entire path, instead of the name to avoid that.

src/javascript/Edit/Edit.jsx
```diff
@@ -39,13 +46,31 @@ export const EditCmp = ({
...
+                        }
+
+                        setStoredLocation({
+                            ...storedLocation,
+                            location: {
+                                ...storedLocation.location,
+                                pathname: storedLocation.location.pathname.replace(node.name, newName)
```
jkevan, line 71: Dangerous, you may rename an element of the jcontent that should not be renamed, see if you can export jcontent fct in the registry to be able to extract jcontent path info from the current location and replace the path to rebuild the jcontent url.

jkevan: Take a closer look this two fcts:
**extractParamsFromUrl** and **buildUrl** from **JContent.redux.js**

src/javascript/Edit/Edit.jsx
```diff
@@ -39,13 +46,31 @@ export const EditCmp = ({
...
+                            }
+                        });
```
jkevan, line 73: This code is only related to redux, please add this in the envProps and env mechanism that is proper to redux or standalone mode.
Look at ContentEditor.redux.jsx to initialize such environmental stuff

src/javascript/ContentEditor.redux.jsx
```diff
@@ -17,18 +20,42 @@ const mapStateToProps = state => {
...
     }
 
+    const rename = (node, mutateNode) => {
+        if (mutateNode && mutateNode.rename) {
+            if (storedLocation && storedLocation.location) {
+                if (openPaths) {
+                    dispatch(replaceOpenedPath(openPaths.map(openPath => openPath.replace(node.path, mutateNode.rename))));
+                }
+
+                const {site: currentSite, language, mode: currentMode, path} = registry.get('jcontent', 'utils').extractParamsFromUrl(storedLocation.location.pathname, storedLocation.location.search);
```
jkevan, line 39: ### ISSUE 1:
you pass the storedLocation.location.search but you never get the associated params, so the rebuilt url will not have the original search querystring if there is one.
Scenario:
- open Jcontent
- search for a content
- edit the current folder and rename
- do back
- you should get back to the renamed folder with the previous search (currently the search is lost)

src/javascript/Edit/Edit.jsx
```diff
@@ -39,13 +40,13 @@ export const EditCmp = ({
                 sections,
                 values
             },
-            editCallback: nodeUuid => {
+            editCallback: (node, mutateNode) => {
                 if (values[Constants.systemFields.OVERRIDE_SUBMIT_CALLBACK]) {
-                    values[Constants.systemFields.OVERRIDE_SUBMIT_CALLBACK]();
+                    values[Constants.systemFields.OVERRIDE_SUBMIT_CALLBACK](node, mutateNode);
```
jkevan, line 45: ### ISSUE 2:

We have an issue here:
The language switcher will not perform the renaming operation.
because it's overriding the submit callback to switch the langue, but it should also handle the rename.
Scenario:
- open jcontent
- edit Home page
- rename system name
-switch language: confirmation popup: save changes (no rename handling)
- go back
- jcontent issue, because the rename have not been handled.

So I think we should always do:
envProps.rename(node, mutateNode);
values[Constants.systemFields.OVERRIDE_SUBMIT_CALLBACK](node, mutateNode);

With this, we insure that no matter the override submit callback we always handle the rename on the save.
I thinks it will be even better if we always do it on top of the editCallback, no matter if it's an overrided or not, like: 

> envProps.rename(node, mutateNode);
> if (values[Constants.systemFields.OVERRIDE_SUBMIT_CALLBACK]) {
>____values[Constants.systemFields.OVERRIDE_SUBMIT_CALLBACK](node, mutateNode);
> } else {
> ____const envEditCallback = envEditCallbacks[contentEditorConfigContext.env];
> ____if (envEditCallback) {
> ________envEditCallback(node, mutateNode, contentEditorConfigContext);
> ____}
> }

So we insure the rename is always handled no matter the case of saving.
And we avoid to have to handle it specifically for every kind of save operations
do not forget to remove it from the **envEditCallback**

src/javascript/Edit/Edit.env.js
```diff
@@ -1,8 +1,12 @@
...
 const envEditCallbacks = {};
-envEditCallbacks[Constants.env.standalone] = (nodeUuid, contentEditorConfigContext) => {
-    contentEditorConfigContext.envProps.editCallback(nodeUuid);
+envEditCallbacks[Constants.env.standalone] = (node, mutateNode, contentEditorConfigContext) => {
+    contentEditorConfigContext.envProps.editCallback(node.uuid);
+};
+
+envEditCallbacks[Constants.env.redux] = (node, mutateNode, contentEditorConfigContext) => {
+    contentEditorConfigContext.envProps.renameNode(node, mutateNode);
 };
```
jkevan, line 10: can be removed if you do:  **ISSUE 2**

src/javascript/ContentEditor.redux.jsx
```diff
@@ -38,7 +65,8 @@ const ContentEditorReduxCmp = ({client, mode, uuid, lang, uilang, site, contentT
         unregisterListeners: () => {
             unRegisterBlockListener();
         },
-        shouldRedirectBeadCrumb: () => false
+        shouldRedirectBeadCrumb: () => false,
+        renameNode: rename
```
jkevan, line 69: rename the **renameNode** by something more meaningful like: **handleRenamedNode** or something else

PR #814, opened 2020-08-11
BACKLOG-14293: avoid using a view for the zoom preview wrapper, inste…

jkevan:


src/main/java/org/jahia/modules/contenteditor/render/PreviewWrapperFilter.java
```diff
@@ -58,6 +58,8 @@
 public class PreviewWrapperFilter extends AbstractFilter {
 
     private final String CE_PREVIEW_WRAPPER = "ce_preview_wrapper";
```
dgriffon, line 60: is that variable still used?

PR #817, opened 2020-08-11
BACKLOG-14392 : add direct parent in breadcrumb

jsinovassin:
## JIRA

https://jira.jahia.org/browse/BACKLOG-14392
## Tests

- [ ] Integration Tests

## Checklist

- [ ] Performance
- [ ] Migration
- [ ] Code maintainability

## Documentation

- [ ] Inline documentation


src/javascript/EditPanel/header/ContentBreadcrumb/ContentPath/ContentPath.container.jsx
```diff
@@ -20,10 +21,18 @@ const findLastIndex = (array, callback) => {
     return lastIndex;
 };
 
-const getItems = (node = {}) => {
-    const ancestors = node.ancestors || [];
+const getItems = (mode, node = {}) => {
+    let ancestors = node.ancestors || [];
 
-    if ((ancestors.length === 0) || node.isVisibleInContentTree) {
+    if ((ancestors.length === 0)) {
```
jkevan, line 27: if we are in create mode, and there is no ancestors, at lest there is the current node that should be considered as an ancestors, normally.

PR #826, opened 2020-08-17
BACKLOG-14374 : name of attribute ref is changed to setOpenRef so nee…

Yousria:
…d to be updated in content-editor to make editing actions work

## JIRA

https://jira.jahia.org/browse/BACKLOG-14374

## Description

 name of attribute ref is changed to setOpenRef so it needs to be updated in content-editor to make editing actions work

### DIFF
--- a/src/javascript/EditPanel/EditPanelContent/FormBuilder/Section/FieldSet/Field/Field.actions.jsx
+++ b/src/javascript/EditPanel/EditPanelContent/FormBuilder/Section/FieldSet/Field/Field.actions.jsx
@@ -31,7 +31,7 @@ export const FieldActionsCmp = ({classes, inputContext, selectorType, field, for
         } else {
             actionCmp = (
                 <>
-                    <ContextualMenu ref={contextualMenu}
+                    <ContextualMenu setOpenRef={contextualMenu}
                                     actionKey={selectorType.key + 'Menu'}
                                     context={{}}
                     />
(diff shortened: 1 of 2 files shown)


src/javascript/EditPanel/EditPanelContent/FormBuilder/Section/FieldSet/Field/Field.actions.jsx
```diff
@@ -31,7 +31,7 @@ export const FieldActionsCmp = ({classes, inputContext, selectorType, field, for
         } else {
             actionCmp = (
                 <>
-                    <ContextualMenu ref={contextualMenu}
```
tdraier, line 34: Not only the name but also the content which changed. All usages of contextualMenu should be updated

Yousria: do you mean the value of setOpenRef (if yes, it's a ref as used in jcontent .. ?) ? or another attribute ?

tdraier: just mean that if it's a ref to open, you cannot call contextualMenu.current.open(), but directly contextualMenu.current(). ContextualMenu has been converted to functional component, so you cannot have a ref to it.

PR #831, opened 2020-08-19
BACKLOG-14442 : handle special characters in system name and lowercase

jsinovassin:
## JIRA

https://jira.jahia.org/browse/BACKLOG-14442

## Tests

### DIFF
--- a/src/javascript/SelectorTypes/SystemName/SystemName.utils.jsx
+++ b/src/javascript/SelectorTypes/SystemName/SystemName.utils.jsx
@@ -2,3 +2,52 @@ export const limitSystemNameIfNecessary = (currentValue, field) => {
     const maxLength = field.selectorOptions.find(option => option.name === 'maxLength')?.value;
     return maxLength ? currentValue.substring(0, maxLength) : currentValue;
 };
+
+const mapSpecialCharacter = {
+    é: 'e',
+    è: 'e',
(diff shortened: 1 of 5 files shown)


dgriffon (approved):

src/javascript/SelectorTypes/SystemName/SystemName.utils.jsx
```diff
@@ -2,3 +2,52 @@ export const limitSystemNameIfNecessary = (currentValue, field) => {
...
+    î: 'i',
+    ï: 'i',
+    Î: 'i',
+    Í: 'i',
+    Œ: 'oe',
+    Æ: 'ae'
+};
+
+export const replaceSpecialCharacters = systemName => {
+    const newSystemName = systemName?.replace(/[^A-Z0-9-_,()'!]/ig, character => {
```
Taybou, line 46: For me these: `,()'!` are special characters we should replace it

Taybou (approved):

src/javascript/SelectorTypes/SystemName/SystemName.utils.jsx
jsinovassin: I Check with romain, it's not necessary to replace these characters because they don't causes problem when they are in a URL

PR #834, opened 2020-08-20
BACKLOG-14443: move systemname to top for jmix:mainResource nodes

jkevan:


src/javascript/Create/CreateForm/createForm.gql-queries.js
```diff
@@ -91,6 +91,14 @@ export const FormQuery = gql`
             nodeTypeByName(name: $primaryNodeType) {
                 name
                 displayName(language: $uilang)
+                moveSystemNameToTop: isNodeType(type: {multi: ANY, types: [
```
dgriffon, line 94: Can you move the list of types to a constant (as it is the same for create and edit) ? 
Thank you

dgriffon (approved):

PR #837, opened 2020-08-27
BACKLOG-14500 support skip filter for readOnly field

Taybou:
## JIRA

https://jira.jahia.org/browse/BACKLOG-14500

## Tests

- [ ] Integration Tests

## Checklist

- [ ] Performance
- [ ] Migration
- [ ] Code maintainability

## Documentation

- [ ] Inline documentation


dgriffon (changes requested):
We won't add such specific case to support that behavior, we need to provide such support at field level like supporting disabled / read only that would help to handle such case.

PR #842, opened 2020-09-07
BACKLOG-14592 update DisplayActions context

Taybou:
https://jira.jahia.org/browse/BACKLOG-14592

### DIFF
--- a/src/javascript/Create/CreateForm/create.action.js
+++ b/src/javascript/Create/CreateForm/create.action.js
@@ -4,7 +4,7 @@ import React, {useContext, useState} from 'react';
 import {ComponentRendererContext} from '@jahia/ui-extender';
 import * as PropTypes from 'prop-types';
 
-const Create = ({context, render: Render, loading: Loading}) => {
+const Create = ({context, values, errors, dirty, render: Render, loading: Loading}) => {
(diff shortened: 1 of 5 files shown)


dgriffon (changes requested):
I am not sure only changing the displayAction to remove the context should be enough.. are you sure component that were using those actions placeholder didn't look for the values of the context and not in the props?

Taybou:
The component looks to this prop:

```
                icon={context.buttonIcon}
                label={t(context.buttonLabel).toUpperCase()}
                disabled={context.disabled}
                data-sel-role={context.dataSelRole}
                onClick={e => {
                    e.stopPropagation();
                    context.onClick(context, e);
                }}
```

Otherwise, what did you suggest to do?

dgriffon:
```icon={context.buttonIcon}
                label={t(context.buttonLabel).toUpperCase()}
                disabled={context.disabled}
                data-sel-role={context.dataSelRole}
                onClick={e => {
                    e.stopPropagation();
                    context.onClick(context, e);
                }}```
I mean that we need to change all `context.*` usages to use the prop directly, adding it to the function signature and PropTypes.

dgriffon (approved):

PR #843, opened 2020-09-08
BACKLOG-14599 : add onCreate / onEdit hooks

dgriffon:
https://jira.jahia.org/browse/BACKLOG-14599

### DIFF
--- a/src/javascript/Edit/save/save.request.js
+++ b/src/javascript/Edit/save/save.request.js
@@ -38,7 +38,7 @@ export const saveNode = ({
         wipInfo
     });
 
-    // Hooks on content to be created
+    // Hooks on content to be edited
     const onEdits = registry.find({type: 'contentEditor.onEdit'});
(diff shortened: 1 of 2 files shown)


src/javascript/Edit/save/save.request.js
```diff
@@ -25,19 +26,31 @@ export const saveNode = ({
...
+        shouldModifyChildren,
+        childrenOrder,
+        wipInfo
+    });
+
+    // Hooks on content to be created
+    const onEdits = registry.find({type: 'contentEditor.onEdit'});
+    variables = onEdits?.reduce((updatedVariables, registeredOnEdit) => {
+        try {
+            return registeredOnEdit.onCreate(updatedVariables, nodeData) || updatedVariables;
```
jsinovassin, line 45: It would be better to name the function onEdit instead of onCreate

dgriffon: Thanks a lot for the comment, name updated!

jsinovassin (approved):

PR #857, opened 2020-09-22
BACKLOG-14498 : fix title and system name comparison to hide copy button

jsinovassin:
https://jira.jahia.org/browse/BACKLOG-14498

### DIFF
--- a/src/javascript/SelectorTypes/SystemName/SystemName.jsx
+++ b/src/javascript/SelectorTypes/SystemName/SystemName.jsx
@@ -3,7 +3,7 @@ import React from 'react';
 import {FieldPropTypes} from '~/FormDefinitions/FormData.proptypes';
 import Text from '~/SelectorTypes/Text';
 import {Constants} from '~/ContentEditor.constants';
-import {limitSystemNameIfNecessary, replaceSpecialCharacters} from './SystemName.utils';
+import {limitSystemNameIfNecessary, replaceSpecialCharacters, isEqualToSystemName} from './SystemName.utils';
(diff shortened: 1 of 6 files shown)


src/javascript/SelectorTypes/SystemName/SystemName.jsx
```diff
@@ -11,6 +11,7 @@ import classes from './SystemName.scss';
 export const SystemNameCmp = ({field, value, values, id, editorContext, onChange}) => {
     const {t} = useTranslation();
 
+    const cleanedTitle = limitSystemNameIfNecessary(replaceSpecialCharacters(values['jcr:title'])?.toLowerCase(), field);
```
jkevan, line 14: This will be executed all the time adding extra calculation something when it's not necessary, best practice would be to put this into a function to be able to call it when necessary only.
from the isDisabled and the onClick.

Or at least use the same condition as: values['jcr:title'] !== undefined && editorContext.mode === Constants.routes.baseEditRoute

src/main/resources/javascript/locales/en.json
```diff
@@ -209,7 +209,7 @@
           "system": {
             "displayName": "System",
             "fields": {
-              "systemNameDescription": "{{maxNameSize}} characters maximum, special characters will be ignored. Used to build the URL for pages if no vanity URL is set in advanced options.",
```
jkevan, line 212: What about other i18n files ? I can see they still mention what you have deleted here for "en".

jkevan (changes requested):

jkevan (approved):

PR #864, opened 2020-10-08
QA-13158 handle case when reference not found & locales

Taybou:
https://jira.jahia.org/browse/QA-13158

### DIFF
--- a/src/javascript/SelectorTypes/Picker/MediaPicker/index.spec.js
+++ b/src/javascript/SelectorTypes/Picker/MediaPicker/index.spec.js
@@ -20,17 +20,22 @@ describe('MediaPicker config', () => {
 
         it('should return no data, no error when loading', () => {
             setQueryResponseMock({loading: true});
-            expect(usePickerInputData('uuid', {lang: 'fr'})).toEqual({loading: true});
+            expect(usePickerInputData('uuid', {lang: 'fr'})).toEqual({loading: true, notFound: true});
         });
 
         it('should return no data when there is no uuid given', () => {
             setQueryResponseMock({loading: false, data: {}});
-            expect(usePickerInputData('', {lang: 'fr'})).toEqual({loading: false});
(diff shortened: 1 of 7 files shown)


src/javascript/SelectorTypes/Picker/MediaPicker/index.spec.js
```diff
@@ -20,17 +20,17 @@ describe('MediaPicker config', () => {
...
         it('should return no data when there is no uuid given', () => {
             setQueryResponseMock({loading: false, data: {}});
-            expect(usePickerInputData('', {lang: 'fr'})).toEqual({loading: false});
+            expect(usePickerInputData('', {lang: 'fr'})).toEqual({loading: false, notFound: false});
         });
 
         it('should return error when there is error', () => {
             setQueryResponseMock({loading: false, error: 'oops'});
-            expect(usePickerInputData('uuid', {lang: 'fr'})).toEqual({loading: false, error: 'oops'});
+            expect(usePickerInputData('uuid', {lang: 'fr'})).toEqual({loading: false, error: 'oops', notFound: true});
```
jkevan, line 33: if there is only this scenarios possible.
I dont understand why we add new prop: notFound.
notFound is true when loading or there is an error, so testing for (error || loading) should be enough.

Reading this scenarios:
(loading || error) === notFound
And in that case notFound is useless because we already have loading and error available.

is it missing a scenario case where (loading || error) !== notFound ?

Taybou: we have weird use case with refetch of query, that's why I introduced this notFound option, it will not impact the exiting elements.

jkevan: is the weird use case introducing a case where: (loading || error) !== notFound ?
if yes we are missing a scenario test case for this one.

jkevan (approved):

PR #867, opened 2020-10-14
Backlog 14847 remove context from actions

jsinovassin:
https://jira.jahia.org/browse/BACKLOG-14847


dgriffon (changes requested):
Please don't pass all {{props}} to the nested component, use {{...otherProps}} to filter out props that are used only for the component.

dgriffon (approved):

PR #875, opened 2020-11-03
BACKLOG-14948 style header of date picker

Taybou:
https://jira.jahia.org/browse/BACKLOG-14948


src/javascript/DesignSystem/DatePicker/YearMonthSelector/YearMonthSelector.jsx
```diff
@@ -19,31 +37,48 @@ const YearMonthSelectorCmp = ({date, months, onChange}) => {
...
 
     const handleChange = e => {
-        const {year, month} = e.target.form;
-        onChange(new Date(year.value, month.value));
+        const {name, value} = e.target;
+        if (name === 'month') {
+            onChange(new Date(selectedYear, value));
+        } else {
+            onChange(new Date(value, selectedMonth));
+        }
```
jkevan, line 45: Could be great to separate this in two different handleChange. now that we handle them outside of the form, also we do not need the name from targe if we do that.

- handleMonthChange
- handleYearChange

jkevan (changes requested):

jkevan (approved):

PR #882, opened 2020-11-19
BACKLOG-15015 : add missing props due to action framework refactoring

dgriffon:
https://jira.jahia.org/browse/BACKLOG-15015


Taybou (approved):

Taybou:
should we cover this new props with unit tests, wdyt?

dgriffon:
> should we cover this new props with unit tests, wdyt?

This is not really a new prop, they were passed in the context object previously, and used by a private function (onclick) 
Test is always good to have but in that case, it does not change the behaviour of the component.

PR #883, opened 2020-11-20
QA-13375 handle fallback selector type in case of unknown

Taybou:
https://jira.jahia.org/browse/QA-13375

### DIFF
--- a/src/javascript/SelectorTypes/SelectorTypes.js
+++ b/src/javascript/SelectorTypes/SelectorTypes.js
@@ -77,7 +77,7 @@ export const registerSelectorTypes = ceRegistry => {
     registerSystemNameOnChange(ceRegistry);
 };
 
-export const resolveSelectorType = ({selectorType, selectorOptions}) => {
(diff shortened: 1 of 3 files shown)


src/javascript/SelectorTypes/SelectorTypes.js
```diff
@@ -87,4 +87,12 @@ export const resolveSelectorType = ({selectorType, selectorOptions}) => {
         selector.key = selectorType;
         return selector;
     }
+
+    if (selectorType) {
+        console.warn(`No renderer component for ${selectorType} selectorType`);
+    } else {
+        console.error(`Field ${displayName} has no selectorType !`, field);
```
dgriffon, line 94: The `field` used in here is missing the `selectorOptions` (`selectorType` is undefined and `displayName` in the previous sentence), we can add it to the log.

dgriffon (approved):

src/javascript/SelectorTypes/SelectorTypes.js
Taybou: ok

PR #887, opened 2020-11-23
QA-13450 : allow to have same name propety in dynamic choicelists

dgriffon:
https://jira.jahia.org/browse/QA-13450

### DIFF
--- a/src/main/java/org/jahia/modules/contenteditor/api/forms/impl/EditorFormServiceImpl.java
+++ b/src/main/java/org/jahia/modules/contenteditor/api/forms/impl/EditorFormServiceImpl.java
@@ -432,7 +432,9 @@ public class EditorFormServiceImpl implements EditorFormService {
             EditorFormField editorFormField = generateEditorFormField(itemDefinition, existingNode, parentNode, uiLocale, locale, rank);
 
             editorFormFields.add(editorFormField);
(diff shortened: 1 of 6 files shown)


src/main/java/org/jahia/modules/contenteditor/api/forms/impl/EditorFormServiceImpl.java
```diff
@@ -432,7 +432,9 @@ private EditorFormFieldSet generateEditorFormFieldSet(Set<String> processedPrope
             EditorFormField editorFormField = generateEditorFormField(itemDefinition, existingNode, parentNode, uiLocale, locale, rank);
 
             editorFormFields.add(editorFormField);
-            processedProperties.add(itemDefinition.getName());
+            if (!dynamic) {
```
Taybou, line 435: so you will only add not dynamic fieldset (mixin), right?

Taybou (approved):

jkevan (approved):

src/main/java/org/jahia/modules/contenteditor/api/forms/impl/EditorFormServiceImpl.java
dgriffon: yes, that's the point to allow several mixins to add the same property.

PR #915, opened 2021-01-11
BACKGLOG-15294 : move picker config to registry

jsinovassin:
https://jira.jahia.org/browse/BACKLOG-15294


src/javascript/SelectorTypes/Picker/Picker.configs.js
```diff
@@ -74,111 +74,147 @@ const treeConfigs = {
...
-    },
-    _getPickerType(options) {
-        const option = options && options.find(option => option.name === 'type' && this[option.value]);
-        return (option && option.value) || 'editorial';
+const _getPickerType = options => {
+    const pickerOption = options && options.find(option => option.name === 'type' && registry.get('pickerConfiguration', option.value));
+    return (pickerOption && pickerOption.value) || 'editorial';
+};
+
+export const resolveConfig = (options, field) => {
```
dgriffon, line 83: We could have move `resolveConfig` and `getPickerSelectorType` to the utility module and keep that one as the registering of the picker configs.

src/javascript/SelectorTypes/Picker/Picker.configs.js
```diff
@@ -74,111 +74,147 @@ const treeConfigs = {
...
+
+    return config;
 };
 
-export default pickerConfigs;
+export const getPickerSelectorType = options => {
+    return registry.get('pickerConfiguration', _getPickerType(options)).cmp.picker;
+};
+
+export const registerPickerConfig = ceRegistry => {
```
dgriffon, line 99: I see that we put all selectors at the root of the registry with global names, it would have be better to namespace them for Content editor.

dgriffon (changes requested):

dgriffon (approved):

PR #916, opened 2021-01-13
BACKLOG-15224 : unregister upload image refetch

dgriffon:
https://jira.jahia.org/browse/BACKLOG-15224


XiangYuHe (commented):
Probably a good idea to not call registry.registry directly 
As we do have a class to house the functionalities

XiangYuHe (approved):
lgtm

PR #927, opened 2021-01-18
QA-13326 : Improve picker labels

jsinovassin:
https://jira.jahia.org/browse/QA-13326

### DIFF
--- a/src/javascript/SelectorTypes/Picker/Picker.configs.js
+++ b/src/javascript/SelectorTypes/Picker/Picker.configs.js
@@ -1,4 +1,3 @@
-
 const treeConfigs = {
     content: {
         rootPath: site => `/sites/${site}/contents`,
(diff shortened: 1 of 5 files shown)


src/javascript/SelectorTypes/Picker/Picker.configs.js
```diff
@@ -108,7 +107,17 @@ export const registerPickerConfig = ceRegistry => {
 
     ceRegistry.add('pickerConfiguration', 'folder', {
         cmp: {
-            picker: contentPicker,
+            picker: {
```
Taybou, line 110: why not update directly the source of "contentPicker"

Taybou (commented):
.

src/javascript/SelectorTypes/Picker/Picker.configs.js
jsinovassin: Because, we don't want to change the label for every type

Taybou (approved):

PR #935, opened 2021-01-27
BACKLOG-15356 Expose tag

AKarmanov:
## JIRA

https://jira.jahia.org/browse/BACKLOG-15356

## Description

Expose Tag to be able to override j:tagList field in a different module while keeping original look and feel

### DIFF
--- a/webpack.config.js
+++ b/webpack.config.js
@@ -100,7 +100,8 @@ module.exports = (env, argv) => {
                 library: { type: "assign", name: "appShell.remotes.contentEditor" },
                 filename: "remoteEntry.js",
                 exposes: {
-                    './init': './src/javascript/init'
+                    './init': './src/javascript/init',
+                    './Tag': './src/javascript/SelectorTypes/Tag/Tag.jsx'
                 },
                 remotes: {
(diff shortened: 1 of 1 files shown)


jkevan (commented):
using the registry to get the component is not enough ?

ceRegistry.add('selectorType', 'Tag', {cmp: Tag, supportMultiple: true});

you could get the original component Tag from registry.get(....).cmp
Then you could create your own selector type component that would wrap the existing component and provide the auto tag button.

AKarmanov:
> using the registry to get the component is not enough ?
> 
> ceRegistry.add('selectorType', 'Tag', {cmp: Tag, supportMultiple: true});
> 
> you could get the original component Tag from registry.get(....).cmp
> Then you could create your own selector type component that would wrap the existing component and provide the auto tag button.
> 
> (we are doing it for the SystemName selector type, it's just wrapping an existing Text selector type component and providing a sync button to sync with jcr:title, I imagine you could do the same, but instead of declaring the Tag component in you code you can get it from the registry)

Yes, indeed this should work. I didn't see that it was also registered in the registry. Thanks!

PR #944, opened 2021-04-01
BACKLOG-14457: Display fieldset description

gflores-jahia:
## JIRA

https://jira.jahia.org/browse/BACKLOG-14457

## Description

Display Fieldset description value in content editor:

image

### DIFF
--- a/src/javascript/EditPanel/EditPanelContent/FormBuilder/FieldSet/FieldSet.jsx
+++ b/src/javascript/EditPanel/EditPanelContent/FormBuilder/FieldSet/FieldSet.jsx
@@ -18,8 +18,15 @@ let styles = theme => ({
     fieldsetTitleContainer: {
         borderTop: `1px solid ${theme.palette.ui.omega}`,
         display: 'flex',
-        flexDirection: 'row',
+        flexFlow: 'row wrap',
+        alignItems: 'center',
+        minHeight: '74px',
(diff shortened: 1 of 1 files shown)


jsinovassin (commented):
The build is failing due to an error memory when executing the javascript test:
Failed to run task: 'yarn build:production' failed. org.apache.commons.exec.ExecuteException: Process exited with an error: 137

It's strange because we already allocated a lot of memory to the build job:
https://github.com/Jahia/content-editor/blob/66eb81c1fe6f1bc1cd028b287a0798ede148b1a4/.circleci/config.yml#L34
If the issue appears again we could use 2xlarge instead of xlarge, or find a way to launch the tests without requiring a lot of memory.

I tried to rerun the workflow and the build pass successfully.

By the way we could rewrite the circleci file to make the build works without the github action like the others projects.

src/javascript/EditPanel/EditPanelContent/FormBuilder/FieldSet/FieldSet.jsx
```diff
@@ -18,8 +18,15 @@ let styles = theme => ({
     fieldsetTitleContainer: {
```
jsinovassin, line 18: (Not mandatory) You can use a scss file to allow to remove the HOC "withStyles" to set the style.
We did it for some others components.

jsinovassin (approved):

PR #964, opened 2021-05-11
Backlog 15829 placeholder

AKarmanov:
## JIRA

https://jira.jahia.org/browse/BACKLOG-15829

## Description

Revert previous changes, allow choicelist initializer to handle $currentSite* placeholders, don't pass parent node to prevent context resolving to /sites which results in incorrect resolution of current site (resolves to system site).

### DIFF
--- a/src/main/java/org/jahia/modules/contenteditor/api/forms/impl/EditorFormServiceImpl.java
+++ b/src/main/java/org/jahia/modules/contenteditor/api/forms/impl/EditorFormServiceImpl.java
@@ -564,11 +564,6 @@ public class EditorFormServiceImpl implements EditorFormService {
         if (selectorOptions != null && !selectorOptions.isEmpty()) {
             Map<String, ChoiceListInitializer> initializers = choiceListInitializerService.getInitializers();
 
-            Map<String, String> replacements = new HashMap<>();
-            if (existingNode != null) {
-                replacements.put("$currentSite", existingNode.getResolveSite().getPath());
-            }
-
             Map<String, Object> context = new HashMap<>();
             context.put("contextType", primaryNodeType);
             context.put("contextNode", existingNode);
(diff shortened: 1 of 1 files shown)


src/main/java/org/jahia/modules/contenteditor/api/forms/impl/EditorFormServiceImpl.java
```diff
@@ -162,7 +162,7 @@ public EditorForm getEditForm(Locale uiLocale, Locale locale, String uuidOrPath)
                     extendContext.put(contextEntry.getKey(), contextEntry.getValue());
                 }
 
-                return getValueConstraints(nodeTypeRegistry.getNodeType(primaryNodeType), editorFormField, node, parentNode, locale, extendContext);
+                return getValueConstraints(nodeTypeRegistry.getNodeType(primaryNodeType), editorFormField, node, locale, extendContext);
```
gflores-jahia, line 165: since we're removing parentNode, maybe we can clean up this line as well: https://github.com/Jahia/content-editor/blob/c2f6aec09a54eb99db4ee768b375885ea2e9af7a/src/main/java/org/jahia/modules/contenteditor/api/forms/impl/EditorFormServiceImpl.java#L152

but I also have a question about removing parent node context below as well so this could also be impacted

src/main/java/org/jahia/modules/contenteditor/api/forms/impl/EditorFormServiceImpl.java
```diff
@@ -564,19 +564,13 @@ private EditorFormField generateEditorFormField(ExtendedItemDefinition itemDefin
...
 
-            Map<String, String> replacements = new HashMap<>();
-            if (existingNode != null) {
-                replacements.put("$currentSite", existingNode.getResolveSite().getPath());
-            }
-
             Map<String, Object> context = new HashMap<>();
             context.put("contextType", primaryNodeType);
             context.put("contextNode", existingNode);
-            context.put("contextParent", parentNode);
```
gflores-jahia, line 575: are we sure it's safe to remove "contextParent"? Asking since I'm seeing a lot of references to `context.get("contextParent")` in core and not sure if all of them are handling null cases of this properly or what other impacts removing this has. @dgriffon any thoughts on this?

src/main/java/org/jahia/modules/contenteditor/api/forms/impl/EditorFormServiceImpl.java
AKarmanov: We can't, it's used by generateEditorFormField method below

src/main/java/org/jahia/modules/contenteditor/api/forms/impl/EditorFormServiceImpl.java
AKarmanov: There is "fallback" strategy in choicelist initializer to use contextNode in case parentNode is not available, in fact in most initializers where this is used contextNode is tried first and contextParent is fallback (I looked at all the ones in the core).

gflores-jahia (approved):

PR #982, opened 2021-08-09
Qa 13804 Don't edit system name in page composer

AKarmanov:
## JIRA

https://jira.jahia.org/browse/QA-13804

## Description

This is a proposed solution to original issue. It remains to be discussed with PO to see if it's acceptable.


AKarmanov:
The solution was found to be unacceptable by PO.

PR #1029, opened 2022-02-01
QA-14020 : change property comparison to only send modified values on…

jsinovassin:
… save
https://jira.jahia.org/browse/QA-14020


dgriffon (approved):
Looks good, a unit test on `_checkIfValuesAreDifferent` can help to ensure to cover all edge cases

PR #1058, opened 2022-04-19
BACKLOG-16827 Add keydown functionality

AKarmanov:
## JIRA

https://jira.jahia.org/browse/BACKLOG-16827

## Description

Keydown functionality for ctrl+s and esc

### DIFF
--- a/src/javascript/ContentEditor.constants.js
+++ b/src/javascript/ContentEditor.constants.js
@@ -74,5 +74,8 @@ export const Constants = {
         hexColorRegexp: /^#([a-f0-9]{3,4}|[a-f0-9]{4}(?:[a-f0-9]{2}){1,2})\b$/i,
         errorCode: 'invalidColor',
         selectorType: 'Color'
+    },
+    keysCodes: {
+        s: 83
     }
 };
--- a/src/javascript/ContentEditor.constants.js
+++ b/src/javascript/ContentEditor.constants.js
@@ -75,7 +75,7 @@ export const Constants = {
         errorCode: 'invalidColor',
         selectorType: 'Color'
     },
-    keysCodes: {
+    keyCodes: {
         s: 83
     }
 };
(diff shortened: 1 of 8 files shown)


src/javascript/ContentEditor.constants.js
```diff
@@ -74,5 +74,8 @@ export const Constants = {
         hexColorRegexp: /^#([a-f0-9]{3,4}|[a-f0-9]{4}(?:[a-f0-9]{2}){1,2})\b$/i,
         errorCode: 'invalidColor',
         selectorType: 'Color'
+    },
+    keyCodes: {
+        s: 83
```
cedmail, line 79: as it is for saving could we rename s to save?

src/javascript/actions/goBack.action.jsx
```diff
@@ -22,6 +23,20 @@ const GoBack = ({render: Render, componentProps, ...otherProps}) => {
 
     const onCloseDialog = useCallback(() => setOpen(false), [setOpen]);
 
+    useKeydownListener(event => {
+        if (event.keyCode === 27) {
```
cedmail, line 27: could we use a constant here keyCodes.close or keyCodes.back?

cedmail (changes requested):

src/javascript/ContentEditor.constants.js
AKarmanov: It's a key code for letter "s", maybe we should keep it this way to avoid confusion in the future? For escape I can put "esc" just like it appears on keyboard.

src/javascript/actions/goBack.action.jsx
AKarmanov: Ah, good catch forgot about this one

AKarmanov:
@cedmail If we want to go with something like "save", "back" etc. then I will need to rename keyCodes object as the name won't make sense anymore. Let me know, what you think.

cedmail (approved):

cedmail (approved):

PR #1069, opened 2022-04-27
QA-14081: Fixed systemname transform

tdraier:
## JIRA

https://jira.jahia.org/browse/QA-14081

### DIFF
--- a/src/javascript/SelectorTypes/SystemName/SystemName.utils.jsx
+++ b/src/javascript/SelectorTypes/SystemName/SystemName.utils.jsx
@@ -4,54 +4,37 @@ export const limitSystemNameIfNecessary = (currentValue, field) => {
 };
 
 const mapSpecialCharacter = {
-    é: 'e',
-    è: 'e',
-    ë: 'e',
-    Ê: 'e',
-    Ë: 'e',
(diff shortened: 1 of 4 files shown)


src/javascript/SelectorTypes/SystemName/SystemName.utils.jsx
```diff
@@ -4,54 +4,36 @@ export const limitSystemNameIfNecessary = (currentValue, field) => {
...
 export const replaceSpecialCharacters = systemName => {
-    const newSystemName = systemName?.replace(/[^A-Z0-9-_,()!]/ig, character => {
-        return mapSpecialCharacter[character] || '-';
-    });
-
-    return newSystemName?.replace(/-+/g, () => {
-        return '-';
-    });
+    if (systemName) {
+        return systemName
```
gflores-jahia, line 24: Is global matching required in this case? Asking because I've had weird behaviour using it before where results toggles between true and false: https://stackoverflow.com/questions/15559897/unexpected-javascript-regexp-behavior

Can be tested if we have strange behavior with same input twice.

gflores-jahia (approved):

PR #1158, opened 2022-06-17
BACKLOG-17230: Fix firing of registeredOnChange on Fields when switch…

cedmail:
…ing language

Add cypress test to validate

## JIRA

https://jira.jahia.org/browse/BACKLOG-17230

## Description

Solution uses a useMemo to on values[field.name] to feed a useEffect to call registeredOnChange with new value 

## Tests

- [X] Integration Tests

## Checklist

- [ ] Performance
- [ ] Migration
- [ ] Code maintainability

## Documentation

- [ ] Inline documentation


src/javascript/editorTabs/EditPanelContent/FormBuilder/Field/Field.jsx
```diff
@@ -88,6 +88,8 @@ export const Field = ({inputContext, idInput, selectorType, field}) => {
 
     const initialValue = useRef(values[field.name]);
 
+    const currentValue = useMemo(() => values[field.name], [values, field.name]);
```
tdraier, line 91: Why do we need a memo here ? Can't we use `values[field.name]` directly in the effect ?

tdraier (approved):

PR #1205, opened 2022-07-28
BACKLOG-19843 Fix cursor, update selection, adjust structured query f…

AKarmanov:
…or contents

## JIRA

https://jira.jahia.org/browse/BACKLOG-19843

## Description
Add cursor for selectable and double clickable rows 
Fix selection reducer 
Update structured query parameters for contents query
Added selection caption, commented out effect which I don't think we need but not 100% sure just yet

### DIFF
--- a/src/javascript/SelectorTypes/Picker/PickerDialog/PickerDialog2.jsx
+++ b/src/javascript/SelectorTypes/Picker/PickerDialog/PickerDialog2.jsx
@@ -3,6 +3,7 @@ import PropTypes from 'prop-types';
 import {Dialog, Slide} from '@material-ui/core';
 import styles from './PickerDialog.scss';
 import {
+    cePickerAddSelection,
     cePickerClearSelection,
     cePickerContextSite,
     cePickerMode,
@@ -50,6 +51,7 @@ export const PickerDialog = ({
     onClose,
     initialSelectedItem,
     editorContext,
+    inputContext,
     pickerConfig,
     accordionItemProps,
     onItemSelection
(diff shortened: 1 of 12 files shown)


src/javascript/SelectorTypes/Picker/PickerDialog/PickerDialog2.jsx
```diff
@@ -162,6 +164,17 @@ export const PickerDialog = ({
         }
     }, [dispatch, editorContext, isOpen, initialSelectedItem, pickerConfig, selectedItem, state, currentFolderInfo]);
 
+    useEffect(() => {
+        if (isOpen && inputContext.actionContext.fieldData) {
```
tdraier, line 168: can't we rely on the "initialSelectedItem" ?

src/javascript/SelectorTypes/Picker/PickerDialog/RightPanel/ContentLayout/ContentTable/ContentTable.jsx
```diff
@@ -113,9 +114,10 @@ export const ContentTable = ({
 
     useEffect(() => {
         if (selection.length > 0) {
-            const toRemove = selection.filter(s => paths.indexOf(s.path) === -1).map(s => s.path);
+            const toRemove = selection.filter(s => paths.indexOf(s.path) === -1);
             if (toRemove.length > 0) {
-                dispatch(reduxActions.removeSelectionAction(toRemove));
+                // TODO why do we have this?
+                // dispatch(reduxActions.removeSelectionAction(toRemove));
```
tdraier, line 120: That was something in jcontent, to remove selection (preview) if an item is not present in the list anymore. But that does not make sense in picker (where selection is persistent)

tdraier (approved):

src/javascript/SelectorTypes/Picker/PickerDialog/PickerDialog2.jsx
AKarmanov: Fileddata comes from usePickerInputData hook which we have in configs, I think it should be pretty reliable. There can be an error when it is not found in which case fieldData is not set.

src/javascript/SelectorTypes/Picker/PickerDialog/RightPanel/ContentLayout/ContentTable/ContentTable.jsx
AKarmanov: Ok, I will remove it then.

PR #1217, opened 2022-08-04
BACKLOG-19852: Fixes for default selection

tdraier:
## JIRA

https://jira.jahia.org/browse/BACKLOG-19852

## Description

Moved the code into a dedicated component that handle selection to avoid the handling of "isOpen" and related lazy queries.
Better handling of default selection on first picker opening, handle site/mode switch separately.


src/javascript/SelectorTypes/Picker/PickerDialog/SelectionHandler.jsx
```diff
@@ -0,0 +1,166 @@
...
+    const nodesInfo = useQuery(GET_PICKER_NODE, {
+        variables: {
+            paths, lang, uilang
+        }
+    });
+
+    useEffect(() => {
+        if (nodesInfo.data) {
+            const nodes = nodesInfo.data.jcr.nodesByPath;
+            dispatch(cePickerSetSelection(nodes.map(n => ({
```
AKarmanov, line 57: Just a note on this section, I saw something similar in picker configs, didn't try the behaviour but it looks like this is dependent on initialSelectedItem which I think is coming from fieldData via usePickerInputData hook in Picker2.jsx so it looks like we already have this data? https://github.com/Jahia/content-editor/blob/bc3ef5bab5ca04687dee1933db1fc498b0a2f401/src/javascript/SelectorTypes/Picker/configs/ContentPickerConfig.js#L21

AKarmanov (approved):

src/javascript/SelectorTypes/Picker/PickerDialog/SelectionHandler.jsx
tdraier: yes it's true that initialSelectedItem is coming from usePickerInputData/fieldData when using Picker2.jsx , but when using from a RichText it directly comes from the path value (which was probably the reason why the picker was taking a path and not a fieldData or a uuid). 
On the other hand, we should maybe reuse these usePickerInputData if the expected result data depends on the config .. ? Right now I'm formatting the data as in ContentPickerConfig, but MediaPickerConfig provides different values (especially for the url)

PR #1243, opened 2022-08-12
BACKLOG-19904: Reset search when opening a picker

cedmail:
## JIRA

https://jira.jahia.org/browse/BACKLOG-19904

## Tests

- [ ] Integration Tests

## Checklist

- [ ] Performance
- [ ] Migration
- [ ] Code maintainability

## Documentation

- [ ] Inline documentation

### DIFF
--- a/src/javascript/SelectorTypes/Picker/Picker2.redux.js
+++ b/src/javascript/SelectorTypes/Picker/Picker2.redux.js
@@ -138,7 +138,14 @@ export const registerPickerReducer = registry => {
         }),
         [cePickerSetSelection]: (state, action) => ({
             ...state,
-            selection: action.payload
+            selection: action.payload,
+            searchPath: '',
+            searchTerms: '',
+            mode: state.mode === Constants.mode.SEARCH ? state.preSearchModeMemo : state.mode,
+            pagination: {
+                ...state.pagination,
+                currentPage: 0
+            }
         }),
         [cePickerAddSelection]: (state, action) => ({
             ...state,


src/javascript/SelectorTypes/Picker/Picker2.redux.js
```diff
@@ -138,7 +138,14 @@ export const registerPickerReducer = registry => {
...
             ...state,
-            selection: action.payload
+            selection: action.payload,
+            searchPath: '',
+            searchTerms: '',
+            mode: state.mode === Constants.mode.SEARCH ? state.preSearchModeMemo : state.mode,
+            pagination: {
+                ...state.pagination,
+                currentPage: 0
+            }
```
tdraier, line 148: i don't understand why resetting search here in cePickerSetSelection - this one is called when opening a picker, to set the previous selection (and so not called when there's no previous selection)
cePickerSetSearchTerm is already called in toggleOpen  ?

cedmail: cePickerSetSelection is called all the time it seems (with []), toggleOpen I did not noticed as in redux all I saw was the call to cePickerKey. The search is not emptied upon opening (maybe an issue with batch?)

tdraier: cePickerKey is called when opening, the batch is called when closing.
I find strange to reset search when calling cePickerSetSelection - this one could be called anytime else than when opening the picker

gflores-jahia: `toggleOpen` is not called when closing editor but `onItemSelection` instead when there is a selection. `onItemSelection` then calls `setDialogOpen` directly to close the dialog. Perhaps `onItemSelection` should call `toggleOpen` instead to trigger the batch actions.

Either that or call the batch actions to clear selection/search term in the else clause instead (when editor is being opened)

tdraier: right, it would be better to call toggleOpen() . We can also reset the searchTerms when both opening and closing , not only closing ?

cedmail: will reset on opening too as it makes

PR #1249, opened 2022-08-16
BACKLOG-19909 Some fixes

AKarmanov:
## JIRA

https://jira.jahia.org/browse/BACKLOG-19909

## Description
A couple of fixes for two issues (when in jcontent):

1) 
    Have site picker with systemsite selected and editiorial with no selection
    Open site picker then editorial picker (editorial picker list is empty as the site is switched to systemsite)
2) 
    Have site picker and editorial picker empty
    Open editorial picker then site picker (site picker is empty because it used editorial picker's path to get sites)


AKarmanov:
There is another issue, will try to address it differently

gflores-jahia (approved):

PR #1253, opened 2022-08-17
BACKLOG-19920: Add provisioning scripts, use jahia-ee image, fix tests for ci

cedmail:


tests/provisioning-manifest-build.yml
```diff
@@ -4,7 +4,7 @@
   username: ${env:NEXUS_USERNAME}
   password: ${env:NEXUS_PASSWORD}
 - installBundle:
-  - 'mvn:org.jahia.modules/qa-module/3.1.0-SNAPSHOT'
+  - url: 'mvn:org.jahia.modules/qa-module/3.1.0-SNAPSHOT'
```
gflores-jahia, line 7: Is `url` needed here? I'm looking at provisioning doc [here](https://github.com/Jahia/jahia-private/tree/master/bundles/provisioning) and not seeing if that's possible or not. 

If needed, maybe we should do the same for snapshot provisioning as well.

cedmail: It's to be inline with the one generated automatically

gflores-jahia (approved):

PR #1303, opened 2022-09-15
BACKLOG-17390: Modify openInTabAction to be working either from picker field or form actions menu

cedmail:
## JIRA

https://jira.jahia.org/browse/BACKLOG-17390

### DIFF
--- a/src/javascript/SelectorTypes/Picker/actions/openInTabAction.js
+++ b/src/javascript/SelectorTypes/Picker/actions/openInTabAction.js
@@ -1,12 +1,31 @@
 import React from 'react';
 import PropTypes from 'prop-types';
 import {Constants} from '~/ContentEditor.constants';
-import {shallowEqual, useSelector} from 'react-redux';
+import {useContentEditorContext} from '~/contexts';
+import {useQuery} from '@apollo/react-hooks';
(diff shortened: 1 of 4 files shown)


src/javascript/SelectorTypes/Picker/actions/openInTabAction.js
```diff
@@ -1,12 +1,31 @@
...
+import {useContentEditorContext} from '~/contexts';
+import {useQuery} from '@apollo/react-hooks';
+import {OpenInTabActionQuery} from '~/SelectorTypes/Picker/actions/openInTabAction.gql-queries';
 
-export const OpenInTabActionComponent = ({render: Render, loading: Loading, uuid, ...others}) => {
-    const {lang} = useSelector(state => ({
-        lang: state.language
-    }), shallowEqual);
+export const OpenInTabActionComponent = ({render: Render, loading: Loading, path, field, inputContext, ...others}) => {
+    const {lang} = useContentEditorContext();
```
gflores-jahia, line 9: if we're just trying to fetch uuid, maybe we should use `useContentEditorConfigContext` instead since it already has both lang and uuid?

cedmail: it's to open content editor on another uuid than the current one

gflores-jahia (approved):

PR #1340, opened 2022-09-29
BACKLOG-20043 Allow dbclick nav on folders, prevent addition of open …

AKarmanov:
…path

## JIRA

https://jira.jahia.org/browse/BACKLOG-20043

## Description

dbclick on folders in structured mode, do not add that path to open paths (so that there is no structure on back navigation)

### DIFF
--- a/src/javascript/SelectorTypes/Picker/PickerDialog/RightPanel/ContentLayout/ContentTable/ContentTable.jsx
+++ b/src/javascript/SelectorTypes/Picker/PickerDialog/RightPanel/ContentLayout/ContentTable/ContentTable.jsx
@@ -78,10 +78,6 @@ export const ContentTable = ({rows, isContentNotFound, totalCount, isLoading, is
         sort: state.contenteditor.picker.sort
     }), shallowEqual);
 
-    const allowDoubleClickNavigation = nodeType => {
-        return Constants.mode.SEARCH !== mode && (['jnt:folder', 'jnt:contentFolder'].indexOf(nodeType) !== -1);
-    };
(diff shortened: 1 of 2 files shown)


src/javascript/SelectorTypes/Picker/PickerDialog/RightPanel/ContentLayout/ContentTable/ContentTable.jsx
```diff
@@ -79,7 +79,7 @@ export const ContentTable = ({rows, isContentNotFound, totalCount, isLoading, is
     }), shallowEqual);
 
     const allowDoubleClickNavigation = nodeType => {
-        return !isStructured && Constants.mode.SEARCH !== mode && (['jnt:folder', 'jnt:contentFolder'].indexOf(nodeType) !== -1);
+        return Constants.mode.SEARCH !== mode && (['jnt:folder', 'jnt:contentFolder'].indexOf(nodeType) !== -1);
```
tdraier, line 82: we will be able to double click on structured view even when tree is not displayed, like in folder picker or editorial link picker , or select folder in page accordion , is this really what we want ?

AKarmanov: Yes, we can navigate folders in those cases as well because of the change and it is probably not what we want to do.

AKarmanov:
@tdraier what do you think of something like this?

tdraier (approved):

PR #1345, opened 2022-09-30
BACKLOG-20046 Initial configuration update

AKarmanov:
## JIRA

https://jira.jahia.org/browse/BACKLOG-20046

## Description

Configuration update to split test execution. DO NOT review until review is requested.


AKarmanov:
So far the split is very simple: picker tests run in standalone mode (on PR) and all tests run in cluster nightly. We don't have too many tests so far and so there aren't many options for splitting.

.github/workflows/on-code-change.yml
```diff
@@ -72,10 +72,11 @@ jobs:
           module_id: content-editor
           testrail_project: Content Editor
           tests_manifest: provisioning-manifest-build.yml
-          jahia_image: jahia/jahia-discovery-dev:8-SNAPSHOT
           should_use_build_artifacts: true
-          ci_startup_script: "ci.startup.sh"
+          jahia_cluster_enabled: false
           should_skip_testrail: true
+          jahia_image: jahia/jahia-ee-dev:8-SNAPSHOT
+          ci_startup_script: "ci.startup.sh"
```
rknj, line 79: Not needed here as it's the same value that is provided by default here: https://github.com/Jahia/jahia-modules-action/blob/main/integration-tests/action.yml#L182 (and used here: https://github.com/Jahia/jahia-modules-action/blob/main/integration-tests/src/docker/start.ts#L21)

rknj (approved):

PR #1346, opened 2022-09-30
BACKLOG-20101: Add page-type filter for content view

gflores-jahia:
## JIRA

https://jira.jahia.org/browse/BACKLOG-20101

## Description

It looks like including `jmix:navMenuItem` to typeFilter also returns `jnt:page` in the results.

Filter out page-like node types when view type is not pages


tdraier (approved):

gflores-jahia:
Merging as I can't reproduce/find issues locally, but will open a separate ticket to investigate.

PR #1367, opened 2022-10-19
BACKLOG-20105 Add read-only badge

AKarmanov:
## JIRA

https://jira.jahia.org/browse/BACKLOG-20105

## Description

Add read only badge


gflores-jahia:
just fyi before we merge, content-editor is in the middle of a 4.1 release so maybe we might need to check first if we'd want to include this one in there or not.

src/javascript/ContentEditor/EditPanel/HeaderBadges/ReadOnlyBadge.jsx
```diff
@@ -0,0 +1,38 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import {Chip, Visibility} from '@jahia/moonstone';
+import {useTranslation} from 'react-i18next';
+import styles from './ReadOnlyBadge.scss';
+import {useContentEditorContext, useContentEditorSectionContext} from '~/contexts';
+
+export const ReadOnlyBadge = ({isGlobal, readOnly}) => {
```
gflores-jahia, line 8: To simplify, I feel like we can get rid of the props here and do the check when we're displaying ReadOnlyBadge instead.

src/javascript/ContentEditor/EditPanel/HeaderBadges/HeaderBadges.jsx
```diff
@@ -4,11 +4,13 @@ import {PublicationInfoBadge} from './PublicationInfoBadge';
 import {LockInfoChip} from './LockInfoChip';
 import {WipInfoChip} from './WipInfoChip';
 import {UnsavedChip} from './UnsavedChip';
+import {ReadOnlyBadge} from './ReadOnlyBadge';
 import {Constants} from '~/ContentEditor.constants';
 import styles from './HeaderBadges.scss';
 
 export const HeaderBadges = ({mode}) => (
     <div className={styles.badges}>
+        {mode === Constants.routes.baseEditRoute && <ReadOnlyBadge isGlobal/>}
```
gflores-jahia, line 13: do the `isGlobal` check here before displaying ReadOnlyBadge

AKarmanov:
> just fyi before we merge, content-editor is in the middle of a 4.1 release so maybe we might need to check first if we'd want to include this one in there or not.

This one is not getting merged until after the release.

src/javascript/ContentEditor/EditPanel/HeaderBadges/ReadOnlyBadge.jsx
AKarmanov: I thought about this, if I delegate isGlobal and the checks on the node and section to a parent component I end up complicating the parent component. Because this badge is also used at field level I would have to do those duplicate checks at field level as well. So I'm not sure it would simplify things, I mean for the badge component itself yes, but everything else badge interacts with would be a bit more complex.

gflores-jahia (approved):

PR #1391, opened 2022-11-04
QA-14194 Update tests

AKarmanov:
## JIRA

https://jira.jahia.org/browse/QA-14194

## Description

Update tests

### DIFF
--- a/tests/cypress/e2e/pickers/constraint.cy.ts
+++ b/tests/cypress/e2e/pickers/constraint.cy.ts
@@ -51,7 +51,7 @@ describe('Picker tests - Constraints', () => {
         const contentEditor = jcontent.createContent('employee');
         const pickerField = contentEditor.getPickerField('qant:employee_supervisor');
         const picker = pickerField.open();
-        cy.wait(200);
+        picker.wait();
         const accordionItem = picker.getAccordionItem('picker-content-folders');
         accordionItem.click();
         picker.wait();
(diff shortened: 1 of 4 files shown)


tests/cypress/e2e/pickers/constraint.cy.ts
```diff
@@ -51,9 +51,11 @@ describe('Picker tests - Constraints', () => {
         const contentEditor = jcontent.createContent('employee');
         const pickerField = contentEditor.getPickerField('qant:employee_supervisor');
         const picker = pickerField.open();
+        cy.wait(200);
```
cedmail, line 54: Do we really need a wait here?

cedmail (changes requested):

tests/cypress/e2e/pickers/constraint.cy.ts
cedmail: from PDF test it seems we don't need
```
const picker = contentEditor.getPickerField('jdmix:fileAttachment_pdfVersion').open();
        picker.getAccordionItem('picker-media').expandTreeItem('images');
```

AKarmanov: I do need a wait but picker.wait() works fine, so I replaced it

cedmail (changes requested):

cedmail (approved):

PR #1395, opened 2022-11-14
BACKLOG-20105 Check every field in section

AKarmanov:
## JIRA

https://jira.jahia.org/browse/BACKLOG-20105

## Description

Check every field in section

### DIFF
--- a/src/javascript/ContentEditor/EditPanel/HeaderBadges/ReadOnlyBadge.jsx
+++ b/src/javascript/ContentEditor/EditPanel/HeaderBadges/ReadOnlyBadge.jsx
@@ -10,7 +10,7 @@ export const ReadOnlyBadge = ({isGlobal, readOnly}) => {
     const {nodeData} = useContentEditorContext();
     const {sections} = useContentEditorSectionContext();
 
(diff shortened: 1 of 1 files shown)


src/javascript/ContentEditor/EditPanel/HeaderBadges/ReadOnlyBadge.jsx
```diff
@@ -10,7 +10,7 @@ export const ReadOnlyBadge = ({isGlobal, readOnly}) => {
     const {nodeData} = useContentEditorContext();
     const {sections} = useContentEditorSectionContext();
 
-    const hasNotReadOnlySection = sections.find(s => s.fieldSets.find(fs => !fs.readOnly));
+    const hasNotReadOnlySection = sections.find(s => s.fieldSets.find(fs => fs.fields.find(f => !f.readOnly)));
```
gflores-jahia, line 13: should we also still keep the check on `fs.readOnly`? I'm not sure if there's a scenario where all fields are not read only but the fieldset itself is set to readOnly?

gflores-jahia (approved):

PR #1408, opened 2022-11-22
BACKLOG-20215 Update css

AKarmanov:
## JIRA

https://jira.jahia.org/browse/BACKLOG-20215

## Description

Update css


cedmail (approved):

gflores-jahia (approved):
For pixel values, should we maybe at least use global spacing vars (at least closest to value)?

PR #1411, opened 2022-11-23
BACKLOG-20309: Truncate path in the middle when too long

cedmail:
## JIRA

https://jira.jahia.org/browse/BACKLOG-20309

## Description

Use overflow/text clipping to keep action button on top
Listen to view port resizing to estimate the size of the path to keep at both end of the string

### DIFF
--- a/src/javascript/ContentEditor/ContentEditor.jsx
+++ b/src/javascript/ContentEditor/ContentEditor.jsx
@@ -6,7 +6,7 @@ import {adaptCreateFormData} from './adaptCreateFormData';
 import {Edit} from './Edit';
 import {EditFormQuery} from './edit.gql-queries';
 import {adaptEditFormData} from './adaptEditFormData';
-import {ContentEditorConfigContextProvider, ContentEditorContextProvider} from '~/contexts';
+import {ContentEditorConfigContextProvider, ContentEditorContextProvider, ViewportProvider} from '~/contexts';
 import {Constants} from '~/ContentEditor.constants';
 import {useFormDefinition} from '~/contexts/ContentEditor/useFormDefinitions';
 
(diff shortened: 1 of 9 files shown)


tdraier:
The solution looks very complex for just a styling issue ? Note that we have the exact the same issue in jcontent, it would be interesting to also fix it there.
From what I have seen, simply adding  `text-overflow: ellipsis; overflow: hidden` on the typography seems enough . Eventually we can also add `min-width: 0;` , but that does not seem necessary.
Also, I've seen that the dom structure is invalid - we have a div inside a span, and actually 2 nested Typography elements - the outer one coming from TableBodyCell, from moonstone, might not be easy to remove ( @ffffffelix ? ) (this has no impact on the overflow)

cedmail:
> The solution looks very complex for just a styling issue ? Note that we have the exact the same issue in jcontent, it would be interesting to also fix it there. From what I have seen, simply adding `text-overflow: ellipsis; overflow: hidden` on the typography seems enough . Eventually we can also add `min-width: 0;` , but that does not seem necessary. Also, I've seen that the dom structure is invalid - we have a div inside a span, and actually 2 nested Typography elements - the outer one coming from TableBodyCell, from moonstone, might not be easy to remove ( @ffffffelix ? ) (this has no impact on the overflow)

I agree that ellipsis could be enough but it is cutting most of the interesting information. another easier solution would be to cut part of the path at the beginning but same issue you might not notice it is in another site

tdraier:
Maybe we can add a tooltip with the full value ? Cutting part at the beginning can be done by using ltr direction , but I don't know if some part is really more interesting than another
Btw just noticed I pushed the test I did on ellipsis here : https://github.com/Jahia/content-editor/pull/1414

cedmail:
Sounds good, merge your PR I will add the title on the Typography

PR #1418, opened 2022-11-23
BACKLOG-20400: Avoid infinite render loop

tdraier:
## JIRA

https://jira.jahia.org/browse/BACKLOG-20400


cedmail (approved):

gflores-jahia (commented):
Maybe just needs some explanation, but I'm not understanding how this causes rendering loop. `defaultCols` and `allColumnData` all looks static arrays in this case. Is it the component references in `allColumnData` items?

tdraier:
defaultCols was not static, it was regenerated on every render (although it could be just generated once). I don't know why it happened only when doing a full refetch (triggered by the job), but the table reducer went into an infinite loop because defaultCols was a new object on every render

PR #1437, opened 2022-11-25
BACKLOG-19878: Removed broken test

tdraier:
## JIRA

https://jira.jahia.org/browse/BACKLOG-19878


gflores-jahia (commented):
from `startWorkflowAction.spec.jsx`, it looks like we lose the checks when form is dirty or node is locked. are we ok with this, or should we replace with explicit checks here? (or at least create a follow-up ticket to add back the checks)

tdraier:
> from `startWorkflowAction.spec.jsx`, it looks like we lose the checks when form is dirty or node is locked. are we ok with this, or should we replace with explicit checks here? (or at least create a follow-up ticket to add back the checks)

I don't see these checks in the tests ? If the form is dirty or locked, the button is disabled, so it cannot be clicked

gflores-jahia:
sorry you're right. I was looking at the associated changeset for when the context check was added. looks like we're covered in those scenarios.

gflores-jahia (approved):

PR #1470, opened 2022-12-07
BACKLOG-20295 : allow to override the mode where the sections are dis…

jsinovassin:
…played

https://jira.jahia.org/browse/BACKLOG-20295

### DIFF
--- a/src/main/java/org/jahia/modules/contenteditor/api/forms/impl/EditorFormServiceImpl.java
+++ b/src/main/java/org/jahia/modules/contenteditor/api/forms/impl/EditorFormServiceImpl.java
@@ -320,7 +320,14 @@ public class EditorFormServiceImpl implements EditorFormService {
 
             EditorFormDefinition mergedFormDefinition = mergeFormDefinitions(formDefinitionsToMerge);
             if (mergedFormDefinition.getSections() != null) {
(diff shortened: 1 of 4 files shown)


src/test/resources/META-INF/jahia-content-editor-forms/forms/nt_base.override.displayModes.json
```diff
@@ -0,0 +1,11 @@
+{
+  "nodeType": "nt:base",
+  "hasPreview": true,
+  "priority": 2.0,
+  "sections": [
+    {
+      "name": "layout",
+      "displayModes": ["edit"]
```
tdraier, line 8: why should we have "layout" section disabled by default for "create" ?

src/main/java/org/jahia/modules/contenteditor/api/forms/impl/EditorFormServiceImpl.java
```diff
@@ -334,6 +341,11 @@ private EditorForm getEditorForm(ExtendedNodeType primaryNodeType, Locale uiLoca
         }
     }
 
+    private boolean shouldHideSection(EditorFormSectionDefinition section, JCRNodeWrapper existingNode){
+        String currentMode = existingNode == null ? "create" : "edit";
```
dgriffon, line 345: Can you add a variable in some way to avoid to do the same test a line 301 `existingNode == null`, then pass the mode directly to the function (instead of the node)

jsinovassin: Done!

PR #1507, opened 2023-01-04
BACKLOG-20536: removed redux state, fixed sections context

tdraier:
## JIRA

https://jira.jahia.org/browse/BACKLOG-20536

### DIFF
--- a/src/javascript/SelectorTypes/registerSelectorTypesOnChange.js
+++ b/src/javascript/SelectorTypes/registerSelectorTypesOnChange.js
@@ -87,7 +87,7 @@ export const registerSelectorTypesOnChange = registry => {
                     }
                 });
                 if (updated) {
-                    onChangeContext.setSections(sections);
+                    onChangeContext.onSectionsUpdate();
                 }
             });
         }
--- a/src/javascript/SelectorTypes/registerSelectorTypesOnChange.js
+++ b/src/javascript/SelectorTypes/registerSelectorTypesOnChange.js
@@ -1,20 +1,19 @@
 import {getFields} from '~/utils/fields.utils';
 import {FieldConstraints} from './registerSelectorTypesOnChange.gql-queries';
 
-function arrayEquals(arr1, arr2) {
-    return arr1.length === arr2.length && arr1.every((value, index) => value === arr2[index]);
-}
-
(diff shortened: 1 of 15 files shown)


src/javascript/SelectorTypes/registerSelectorTypesOnChange.js
```diff
@@ -82,7 +87,7 @@ export const registerSelectorTypesOnChange = registry => {
                     }
                 });
                 if (updated) {
-                    onChangeContext.onSectionsUpdate();
+                    onChangeContext.setSections(sections);
```
jkevan, line 90: This will log the warning message:
`console.warn('Sections update is deprecated');` may be better to keep calling `onSectionsUpdate()` to avoid warn logs.

jkevan (approved):

src/javascript/contexts/ContentEditorSection/ContentEditorSection.context.jsx
```diff
@@ -1,20 +1,39 @@
...
+    const [expanded, setExpanded] = useState({});
     const [, setChangeCount] = useState(0);
 
+    const stringifiedSections = formSections && JSON.stringify(formSections);
+    if (stringifiedSections && previousValue.current !== stringifiedSections) {
+        previousValue.current = stringifiedSections;
+        const sectionsCopy = JSON.parse(stringifiedSections);
+        setSectionsState(sectionsCopy);
+        if (Object.keys(expanded).length === 0) {
+            setExpanded(sectionsCopy.reduce((acc, curr) => ({
```
dgriffon, line 20: `reduce` can be replaced by a `map`

dgriffon (approved):

src/javascript/contexts/ContentEditorSection/ContentEditorSection.context.jsx
tdraier: how ? unless we convert `expanded` to an array (which could be more simple) ?

dgriffon: indeed, the `reduce` is required
```[curr.name]: acc[curr.name] ? acc[curr.name] : curr.expanded``` 
can be simplified

PR #1530, opened 2023-01-27
TECH-1134: Added capability to run manually and across all supported branches

Fgerthoffert:
- Workflow can now be triggered manually
- A matrix job allows the check to be executed on all supported branches

### DIFF
--- a/.github/workflows/schedule-sonar.yml
+++ b/.github/workflows/schedule-sonar.yml
@@ -7,6 +7,9 @@ on:
   schedule:
     - cron: '0 4 * * 1,3,5' # At 4AM, on Monday, Wednesday and Friday
 
+env:
+  MODULE_ID: content-editor
+
 jobs:
   sonar-analysis:
     name: Sonar Analysis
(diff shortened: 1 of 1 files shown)


.github/workflows/schedule-sonar.yml
```diff
@@ -3,20 +3,27 @@
...
 
 jobs:
   sonar-analysis:
-    name: Sonar Analysis
+    name: Sonar Analysis (Branch: ${{ matrix.supported_branches }})
     runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+         supported_branches: ['master', '3_5_x']
```
bpapez, line 17: Who will keep this up-to-date?
It is very likely that it will be forgotten to put new branches into this list.

bpapez (changes requested):

.github/workflows/schedule-sonar.yml
Fgerthoffert: yes, but I'm not sure what the alternative is, scheduled workflows run from the master branch, so at some point we have to specify which branches should be executed (and github has no way of knowing what branches are considered "supported").

bpapez: I can either think of having an automated way to detect which branches are still active: by using a regexp to detect our maintenance branches and check the last commit date and if it is not older than n months we do the sonarqube check. 

Or we should have some standard way to configure per repository (or in a general document/url), which branches are still supported, and then have a workflow/process to make sure that this information is kept up-to-date.

Fgerthoffert: I think we should for the time being go ahead with this. Yes it will rely on someone from the team to think about updating the maintenance branches, but as-is master will be scanned, which is probably better than today

bpapez (approved):

PR #1549, opened 2023-02-13
BACKLOG-17201: Persist 'create another' checkbox state

gflores-jahia:
## JIRA

https://jira.jahia.org/browse/BACKLOG-17201

## Description

Move `createAnother` state to content-editor context to persist when switching languages during create.

### DIFF
--- a/src/javascript/ContentEditor/Create.jsx
+++ b/src/javascript/ContentEditor/Create.jsx
@@ -1,4 +1,4 @@
-import React, {useEffect, useState} from 'react';
+import React, {useEffect} from 'react';
 import {useNotifications} from '@jahia/react-material';
 import {Formik} from 'formik';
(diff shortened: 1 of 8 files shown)


tdraier (dismissed):

cedmail:
Just one question don't we need to keep the create another checked also in non i18n situation

- Site only in english
- Create a Rich Text,
- Click Create Another
- Click Save
- Create Another should still be checked?

cedmail (dismissed):

gflores-jahia:
@cedmail Added basic 'create another' scenario like you described.

Also looks like I was able to fix pdf and pickerMultiple flakiness that's been going on as well, but we'll see if it comes up again on future PRs

cedmail (approved):

PR #1556, opened 2023-02-20
BACKLOG-20565: Allow to use override per site

tdraier:
## JIRA

https://jira.jahia.org/browse/BACKLOG-20565


gflores-jahia (commented):
Code looks ok to me but looks like there's still some JUnit tests failing

gflores-jahia (approved):

PR #1605, opened 2023-04-24
TECH-1143 add datePicker test

Janin-Michel-Mathias:
## JIRA

https://jira.jahia.org/browse/TECH-1143

## Description
Add tests on date picker


Fgerthoffert (changes requested):
As per the discussion on slack:
 - The tests should actually include the test (tests should not be in before or after)
 - `Use utils/helpers/graphql/groovy scripts to do any before/after operation, never load/use the ui here` https://jahia.slack.com/archives/C04JG1LNR9U/p1682069122403599

Fgerthoffert (approved):

PR #1662, opened 2023-07-10
BACKLOG-21296: Ensure we do a deep copy of constraints and that we get same information in both queries

cedmail:
## JIRA

https://jira.jahia.org/browse/BACKLOG-21296
## Description

Issue was in dependent properties constraints being badly compared and missing some properties leading to a chain of refresh emptying the country

## Tests

- [ ] Integration Tests

## Checklist

- [ ] Performance
- [ ] Migration
- [ ] Code maintainability

## Documentation

- [ ] Inline documentation

### DIFF
--- a/src/javascript/SelectorTypes/registerSelectorTypesOnChange.js
+++ b/src/javascript/SelectorTypes/registerSelectorTypesOnChange.js
@@ -2,8 +2,26 @@ import {getFields} from '~/utils/fields.utils';
 import {FieldConstraints} from './registerSelectorTypesOnChange.gql-queries';
 
 function arrayEquals(arr1, arr2) {
-    return arr1.length === arr2.length && arr1.every((value, index) => value === arr2[index]);
+    return arr1.length === arr2.length && arr1.every((value, index) => {
+        if (isObject(value)) {
+            return JSON.stringify(value) === JSON.stringify(arr2[index]);
+        }
+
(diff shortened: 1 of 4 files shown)


src/javascript/SelectorTypes/registerSelectorTypesOnChange.js
```diff
@@ -2,9 +2,27 @@ import {getFields} from '~/utils/fields.utils';
 import {FieldConstraints} from './registerSelectorTypesOnChange.gql-queries';
 
 function arrayEquals(arr1, arr2) {
-    return arr1.length === arr2.length && arr1.every((value, index) => value === arr2[index]);
+    return arr1.length === arr2.length && arr1.every((value, index) => {
+        if (isObject(value)) {
+            return JSON.stringify(value) === JSON.stringify(arr2[index]);
```
gflores-jahia, line 7: this can return false if order of attr is not the same, or at least not guaranteed e.g. `JSON.stringify({x:1, y:2}) ` is not equal to `JSON.stringify({y:2, x:1})`. I'm not sure if it matters here or not.

cedmail: I know but it will not matter here as everything will be in same order

gflores-jahia (approved):

PR #1670, opened 2023-07-12
QA-14376: Add cypress test for default template value, readonly

gflores-jahia:
## JIRA

https://jira.jahia.org/browse/QA-14376

Add priority value in json override to move `j:templateName` field from `jmix:hasTemplateNode` fieldSet to `cent:testDefaultTemplate` to be able to set readOnly


gflores-jahia:
Failing on ```FAILURES:
 | Suite: Mocha Tests - Total tests: 2 - Failure: 1 - Executed in 32s
 |   | - Test list ordering
 |   |    | - FAIL: Test list ordering Activates and deactivates automatic ordering```

which I believe is currently being worked on (also failing on master)

PR #1683, opened 2023-07-24
TECH-1328 add dateTimePicker test

Janin-Michel-Mathias:
## JIRA

https://jira.jahia.org/browse/TECH-1328

## Description
Fix DatePicker test
Add dateTimePicker test


rknj (changes requested):
See my last comment

PR #1707, opened 2024-01-08
TECH-1545 : use webpack shared configuration

dgriffon:
https://jira.jahia.org/browse/TECH-1545


dgriffon:
Won't be done yet in current jahia parent version ( < 8.2.x )

PR #1763, opened 2024-09-30
QA-15296: Allow to reorder fieldset in section

cedmail:
JIRA
https://jira.jahia.org/browse/QA-15296

Description
Allow to reorder fieldset in a section in a JSON form file,and target a fieldset by node type in a JSON fieldset file


gflores-jahia:
Reviewed with @cedmail. We currently have a way to reorder and move fields but only on the fieldset level and not in the section, which is what we're adding here. And it's something that we already have in jcontent 3 so just adding support for the same thing in CE 4.x

gflores-jahia (approved):

PR #1777, opened 2024-12-03
QA-15268: Change preview loading

dgriffon:
https://jira.jahia.org/browse/QA-15268

Change preview loading to use `srcDoc` instead of `document.write`. 
Add a timeout to always remove the loader in case of never ending loading page.


src/javascript/editorTabs/EditPanelContent/Preview/PreviewViewers/IframeViewer/IframeViewer.jsx
```diff
@@ -49,61 +49,50 @@ function loadAssets(assets, iframeDocument) {
...
-        let displayValue = data && data.nodeByPath && data.nodeByPath.renderedContent ? data.nodeByPath.renderedContent.output : '';
-        if (displayValue === '') {
-            displayValue = t('label.contentManager.contentPreview.noViewAvailable');
+    useEffect(() => {
+        // Add a timer to always remove the loader and click on links after 5 seconds
+        onLoadTimeoutRef.current = setTimeout(() => {
+            console.debug('iframe onLoad did not trigger, remove loader and disable click events');
+            const element = iframeRef.current;
+            const iframeWindow = element.contentWindow || element;
+            iframeWindow.document.body.setAttribute('style', 'pointer-events: none');
```
jkevan, line 69: Since the timeout was trigger, we cannot know the state of the iframe.
Do you think we should add undefined protection on this ?
- element ?
- iframeRef.current ?
- iframeWindow.document ? .body ?

src/javascript/editorTabs/EditPanelContent/Preview/PreviewViewers/IframeViewer/IframeViewer.jsx
```diff
@@ -49,61 +49,50 @@ function loadAssets(assets, iframeDocument) {
...
+    const onLoad = () => {
+        try {
+            console.debug('Preview loaded, add assets and remove loader');
+            const element = iframeRef.current;
+            const iframeWindow = element.contentWindow || element;
+            iframeWindow.document.body.setAttribute('style', 'pointer-events: none');
+
+            const assets = data && data.nodeByPath && data.nodeByPath.renderedContent ?
+                data.nodeByPath.renderedContent.staticAssets : [];
+            loadAssets(assets, iframeWindow.document);
```
jkevan, line 84: Is the loading of assets still required when rendering a full page ?

jkevan (dismissed):
Do you think we need memoization of methods like it was done previously ? with usage of `useCallback` ?
I know previous optimizations was done in order to avoid rendering Iframe multiple times due to components props changing.
Just want to make sure we did not introduce regressions on past optimizations.
But it's possible that your version is already optimized, difficult to say by reading the code, it's probably easier to when debugging or analyzing react renders.

jkevan (dismissed):

dgriffon:
> Do you think we need memoization of methods like it was done previously ? with usage of `useCallback` ? I know previous optimizations was done in order to avoid rendering Iframe multiple times due to components props changing. Just want to make sure we did not introduce regressions on past optimizations. But it's possible that your version is already optimized, difficult to say by reading the code, it's probably easier to when debugging or analyzing react renders.

it's not the case with the current implementation as the props sent to the iframe component are always the same, so no refresh happen. But we had an issue when saving displayed content in the preview, the full page is visible before the zoom to happen, it's now fixed using properly the loader. Thank you for pointing this.

jkevan (approved):

PR #1778, opened 2024-12-11
BACKLOG-23452: Provide a way to override date format in date time selector type

dgriffon:
Provide a way to override date format in date time selector type using a configuration name forceDateFormat (DD/MM/YYYY, MM/DD/YYYY are only valid format)

https://jira.jahia.org/browse/BACKLOG-23452

### DIFF
--- a/src/javascript/SelectorTypes/DateTimePicker/DateTimePicker.jsx
+++ b/src/javascript/SelectorTypes/DateTimePicker/DateTimePicker.jsx
@@ -12,15 +12,30 @@ const variantMapper = {
     DateTimePicker: 'datetime'
 };
 
+function getDateFormat(editorContext) {
+    const userNavigatorLocale = editorContext.browserLang;
+    const allowedOverridesDateFormat = ['MM/DD/YYYY', 'DD/MM/YYYY'];
+
+    // Read date format from config
+    const forceDateFormat = window.contextJsParameters?.config?.contentEditor?.forceDateFormat;
+    if (forceDateFormat && !allowedOverridesDateFormat.includes(forceDateFormat)) {
+        console.warn(`forceDateFormat as been set to an invalid value (${forceDateFormat}). Please use one of the following values: ${allowedOverridesDateFormat.join(', ')}`);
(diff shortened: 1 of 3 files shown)


src/javascript/SelectorTypes/DateTimePicker/DateTimePicker.jsx
```diff
@@ -12,15 +12,25 @@ const variantMapper = {
...
 
+// Read global configuration
+const allowedOverridesDateFormat = ['MM/DD/YYYY', 'DD/MM/YYYY'];
+const forceDateFormat = window.contextJsParameters?.config?.contentEditor?.forceDateFormat;
+if (forceDateFormat && !allowedOverridesDateFormat.includes(forceDateFormat)) {
+    console.warn(`forceDateFormat as been set to an invalid value (${forceDateFormat}). Please use one of the following values: ${allowedOverridesDateFormat.join(', ')}`);
+}
+
+const useConfigDateFormatOverride = allowedOverridesDateFormat.includes(forceDateFormat) && forceDateFormat;
+
```
gflores-jahia, line 23: Code looks ok but would prefer to wrap it in a function and call this to set `useConfigDateFormatOverride` variable.

```suggestion
// Read global configuration
function getDateFormatOverride() {
    const allowedOverridesDateFormat = ['MM/DD/YYYY', 'DD/MM/YYYY'];
    const forceDateFormat = window.contextJsParameters?.config?.contentEditor?.forceDateFormat;
    if (forceDateFormat && !allowedOverridesDateFormat.includes(forceDateFormat)) {
        console.warn(`forceDateFormat as been set to an invalid value (${forceDateFormat}). Please use one of the following values: ${allowedOverridesDateFormat.join(', ')}`);
    }
    return allowedOverridesDateFormat.includes(forceDateFormat) && forceDateFormat
}

const useConfigDateFormatOverride = getDateFormatOverride();
```

gflores-jahia (changes requested):

jkevan (approved):

PR #1806, opened 2025-02-28
feat: fix labels in FR and DE

sdusolle:
## JIRA

Fixed some labels in French and German.


gflores-jahia (approved):
Looks ok in terms of json format, but seeing we only have French-Canadian here, maybe @rknj or @Fgerthoffert to verify translation

Fgerthoffert (approved):

